Issue #4, opened 2023-07-26
O3.2.5 Program of work for Moist HS Calibration


odunbar:
```[tasklist]
- [ ] https://github.com/CliMA/Calibratmos.jl/issues/5
- [ ] Add `src/` files for data post-processing of `ClimaAtmos` runs (currently focus on output files from HS) write tests for these in `test/` and with Github actions to run them on Commit, 
- [ ] Add drivers(perhaps in `example/MoistHS`) for running a test case for moist HS , Including an experiment TOML file containing, and example dependencies calibration parameters, a readme documenting some like simulation information e.g. the space/time resolutions, and how long it takes to run at different core counts and the stability at different parameter values.
- [ ] Add files into the `example/MoistHS` file to create and solve an inverse problem. Involving (1) initialize_ensemble.jl to read and sample the prior described in the TOML and to write new TOML samples (2) `process_data.jl` to call the src/ data processing scripts. (3) update_ensemble.jl` to update the ensemble based on the processed data and current ensemble members (4) SLURM scripts to call the pipeline of (1) (2-3) (2-3)... ...(2-3). 
```

nefrathenrici:
Superseded by #17

PR #6, opened 2023-08-01
Set up repo

nefrathenrici:
## Content

----


odunbar (commented):
All the peripherals look great, 

Just two questions from me:
- In the `examples/moist_held_suarez` there is the yml, but where is the "script" to initiate a run? 
- is plot a separate directory to avoid loading Makie etc into src?

Once these are answered/resolved I'd be happy for merge

nefrathenrici:
> In the `examples/moist_held_suarez` there is the yml, but where is the "script" to initiate a run?

I have an initialization script and a single iteration script, but figured it would be better to keep this as a small initial PR and match the plan set out in #4. 

> is `plot` a separate directory to avoid loading Makie etc into src?

Yep, I think it's cleaner to separate it out.

cc @odunbar

odunbar (approved):

PR #41, opened 2024-01-25
Test out precompilation

nefrathenrici:
This PR documents how to force precompilation for a given test case.

In testing, I found that precompiling the model for a given float type did not result in a significant speedup. Instead, precompilation should happen more on a case-by-case basis. We could precompile for the given testcase that is being run, but this only results in a significant improvement in runtime if the compilation time is longer than the runtime, which is fairly rare for calibrations.


nefrathenrici:
Add this to docs with explanation instead of source code

PR #52, opened 2024-02-22
Add a perfect model example using SurfaceFluxes.jl

### LINKED ISSUE #53: Add a perfect model example using SurfaceFluxes.jl
This addresses the following issue of the O3.1.1 Milestone:

https://github.com/CliMA/ClimaCoupler.jl/issues/638

# Components
- [x] implement the SF perfect example, using Businger parameters
- [x] use (and modify where needed) the CalibrateAtmos interface
- [x] show convergence of parameters
- [x] debug insensitivity (better observation data needed)
- [x] ensure the test still runs using the bash scripts
- [ ] adapt Atmos code for the new interface, ensure it runs as before

## Convergence (`a_m` and `a_b` params)
convergence_coefficient_a_h_businger
convergence_coefficient_a_m_businger
scatter_iter

# notes / TODOs
- SF - Why are some Businger params still hard coded? - fixed in  https://github.com/CliMA/SurfaceFluxes.jl/pull/165
- `pipeline.jl` runs without bash scripts and was here for debugging. As part of a separate PR, we can generalize it, so a similar script can be used by other lightweight models. For the heavyweight models, we may be able to wrap the bash / SLURM script execution into Julia's pipeline functions, so we can call the scripts from the model repos.  
- we should use the saved output to retrieve ustar in the plotting script

LenkaNovak:
closes #53 

----
- [x] I have read and checked the items on the review checklist.


docs/src/quickstart.md
```diff
@@ -1,13 +1,28 @@
 # Getting Started
 
+Before you run, make sure your system meets the dependencies of [CalibrateEmulateSample.jl](https://clima.github.io/CalibrateEmulateSample.jl/dev/installation_instructions/). TODO: remove if workaround
+
```
LenkaNovak, line 4: Leaving this here for now.

experiments/surface_fluxes_perfect_model/observation_map.jl
```diff
@@ -0,0 +1,55 @@
...
+Process the data from a single ensemble member to obtain the observation.
+If `output_variance` is true, return the observation and its variance.
+
+This is used to transform the model output to the observation space.
+"""
+
+function process_member_data(ustar; output_variance = false)
+
+    profile_mean = nanmean(ustar)
+    observation = Float64[profile_mean] # Does this have to be FT64?
```
nefrathenrici, line 45: I think this needs to be Float64 to match the FT of the EnsembleKalmanProcess struct

experiments/surface_fluxes_perfect_model/sf_model.jl
```diff
@@ -0,0 +1,142 @@
...
+                ustar_array[ii, kk, ll] = sc.ustar
+            end
+        end
+    end
+
+    # save ustar_array to file
+    JLD2.save(
+        joinpath(model_config["output_dir"], "model_ustar_array.jld2"),
+        Dict("ustar" => ustar_array),
+    )
```
nefrathenrici, line 127: This use `save_object` and `load_object` to save the array directly instead of nesting it within a dict. This isn't a required change, but it is slightly cleaner!
```suggestion
    JLD2.save_object(
        joinpath(model_config["output_dir"], "model_ustar_array.jld2"),
        ustar_array,
    )
```

docs/src/experiment_setup_guide.md
```diff
@@ -39,13 +39,13 @@ First, create a folder for your experiment with a descriptive experiment id in t
...
 
-- be named `atmos_config.yml`. 
+- be named `model_config.yml`.
 - include an `output_dir` to store model output.
 - include the path to the restart file. This is detailed in the next section.
 Here is an example of the arguments you will need:
 ```
 output_dir: output/experiment_name
-restart_file: experiments/experiment_name/restart_file.hdf5 
+restart_file: experiments/experiment_name/restart_file.hdf5
```
akshaysridhar, line 48: ClimaAtmos run restarts are controlled by the `RESTART_FILE` env variable - I don't believe that these are ingested as environment variables based on config dict arguments right now (we set the ENV_VAR in the buildkite pipeline files). cc @nefrathenrici

nefrathenrici (commented):
Thank you Lenka! I really like the new interface. I think we need to deal with the different environments a bit better (mainly plots), but that can be done in the future.

docs/src/experiment_setup_guide.md
nefrathenrici: That is correct, but CalibrateAtmos sets the ENV variable from this config variable - see [here](https://github.com/CliMA/CalibrateAtmos.jl/blob/main/src/atmos_interface.jl#L37).

I have been meaning to set the restart file from the config directly in ClimaAtmos. I will try and find the time this week.

akshaysridhar: Ah - this is internal to the `model_interface` below with the `ENV` variable set within the `get_config` function; no issues then.

docs/src/quickstart.md
```diff
@@ -1,13 +1,30 @@
...
 This experiment runs the Held-Suarez configuration, estimating the parameter `equator_pole_temperature_gradient_wet`.
-By default, it runs 10 ensemble members for 3 iterations. 
+By default, it runs 10 ensemble members for 3 iterations.
 
 To run this experiment:
 1. Log onto the Caltech HPC
 2. Clone CalibrateAtmos.jl and `cd` into the repository.
 3. Run: `bash experiments/pipeline.sh sphere_held_suarez_rhoe_equilmoist 8`. This will run the `sphere_held_suarez_rhoe_equilmoist` experiment with 8 tasks per ensemble member.
 
+## Local Machine
```
akshaysridhar, line 17: Thanks for maintaining and updating the docs!!

experiments/sphere_held_suarez_rhoe_equilmoist/observation_map.jl
```diff
@@ -27,17 +27,17 @@ function latitudinal_avg(arr)
     return dropdims(mean(arr; dims); dims)
 end
 
-function height_avg(arr)
+function height_avg(arr)  # unused
```
akshaysridhar, line 30: Do we want to remove these unused functions entirely? In the future we should be able to use functions defined within ClimaAnalysis (which should also remove the need for hardcoded `dims` info here)

nefrathenrici: These can be removed. I will switch this example over to ClimaAnalysis in a later PR

experiments/surface_fluxes_perfect_model/model_interface.jl
```diff
@@ -0,0 +1,92 @@
...
+end
+
+"""
+    run_forward_model(config::AbstractDict)
+
+Runs the model with the given an AbstractDict object.
+"""
+
+function run_forward_model(::SurfaceFluxModel, config::AbstractDict)
+    FT = Float32
```
akshaysridhar, line 88: Should this be inferred from the float_type of loaded profiles or from the `FLOAT_TYPE` config dict arg in `ClimaAtmos` ?

plot/convergence_surface_fluxes_perfect_model.jl
```diff
@@ -0,0 +1,192 @@
...
+
+            push!(error_vec, ensemble_error)
+            push!(spread_vec, ensemble_spread)
+        end
+
+        u_series = [getindex.(u_vec, i) for i in 1:N_mem]
+        phi_series = [getindex.(phi_vec, i) for i in 1:N_mem]
+
+        f = Makie.Figure(
+            title = "Convergence Plot for $param_name",
```
akshaysridhar, line 60:
```suggestion
```

plot/convergence_surface_fluxes_perfect_model.jl
```diff
@@ -0,0 +1,192 @@
...
+            resolution = (800, 800),
+        ) # title doesn't show up
+
+        ax = Makie.Axis(
+            f[1, 1],
+            xlabel = "Iteration",
```
akshaysridhar, line 66:
```suggestion
            title = "Convergence Plot for $param_name",
            xlabel = "Iteration",
```

experiments/sphere_held_suarez_rhoe_equilmoist/Manifest.toml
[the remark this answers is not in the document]
LenkaNovak: I tested it on the cluster just before the rebase, but I can try to update it to the latest Atmos (by the way, is #61, do we want to change to ClimaParams)?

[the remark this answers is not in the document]
nefrathenrici: Sorry, I meant #55. 
We can update to 20.1 here we'd just need to copy the observations from #55, since those need to be updated as well. It would also be good to `git move` the old experiments Manifest to experiments/sphere_held_suarez_rhoe_equilmoist/Manifest.toml to track package changes.

 It looks like this PR is already switching to ClimaParams, where else do we need to switch?

plot/Manifest.toml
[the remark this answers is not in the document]
LenkaNovak: I included it here, so that the plotting script can be called from the pipeline.jl script, but actually we can just get that example depend on CairoMakie, which seems neater anyway, so I can remove this Manifest.

src/model_interface.jl
```diff
@@ -0,0 +1,62 @@
...
+    error("run_forward_model not implemented for $physical_model")
+
+"""
+    get_physical_model(experiment_id::Val)
+
+Returns the custom physical model objet for the given experiment id. An error is thrown if the experiment id is not recognized.
+"""
+function get_physical_model(experiment_id::Val)
+    error("get_physical_model not implemented for $experiment_id")
+end
```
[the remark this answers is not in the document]
LenkaNovak: That's a great point!

experiments/surface_fluxes_perfect_model/observation_map.jl
LenkaNovak: Ahh, good to know, thanks. I'll add a comment on this and update the SF example.

experiments/surface_fluxes_perfect_model/sf_model.jl
LenkaNovak: Good idea! I've changed this.

test/Manifest_old.toml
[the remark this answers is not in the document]
LenkaNovak: Ahh, this was just a placeholder, well spotted!

experiments/pipeline.jl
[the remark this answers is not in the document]
LenkaNovak: I only put together your already existing pipeline, so thank *you*! 😉

Project.toml
```diff
@@ -5,9 +5,10 @@ version = "0.1.0"
 
 [deps]
 CalibrateEmulateSample = "95e48a1f-0bec-4818-9538-3db4340308e3"
-ClimaAtmos = "b2c96348-7fb7-4fe0-8da9-78d88439e717"
 ClimaComms = "3a4d1b5c-c61d-41fd-a00a-5873ba7a1b0d"
 ClimaCore = "d414da3d-4745-48bb-8d80-42e94e092884"
+ClimaParams = "5c42b081-d73a-476f-9059-fd94b934656c"
+Conda = "8f4d0f93-b110-5947-807f-2305c1781a2d"
```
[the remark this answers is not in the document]
LenkaNovak: That's what was generated by the CalibrateEmulateSample installation, but it looks like we can get away with removing this dependency now! Good catch!

experiments/surface_fluxes_perfect_model/model_interface.jl
LenkaNovak: Yess, that's a great suggestion! I'll make it dependent on the profiles in this example.

experiments/sphere_held_suarez_rhoe_equilmoist/Manifest.toml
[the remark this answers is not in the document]
LenkaNovak: Hmm, it looks like SF.jl needs ClimaParams, whereas ClimaAtmos still needs CLIMAParameters 😅, so I've added the Project of #55 from the experiments to the Atmos example dir, if that's ok. It's not exactly the same Manifest, but it seems to work on the cluster, so lmk if you'd prefer something else.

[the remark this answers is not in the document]
nefrathenrici: That sounds good. The slight issue is that ClimaAtmos is a bit behind, it's only being updated to use ClimaParams now.
Thanks for dealing with this!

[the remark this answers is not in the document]
LenkaNovak: Great, sounds good! We can always update the Manifest of the atmos case later :)

experiments/sphere_held_suarez_rhoe_equilmoist/Project.toml
```diff
@@ -1,13 +1,15 @@
...
 Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
 EnsembleKalmanProcesses = "aa8a2aa5-91d8-4396-bcef-d4f2ec43552d"
+Insolation = "e98cc03f-d57e-4e3c-b70c-8d51efe9e0d8"
 JLD2 = "033835bb-8acc-5ee8-8aae-3f567f8a3819"
 NetCDF = "30363a11-5582-574a-97bb-aa9a979735b9"
 TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
+
+[compat]
+Insolation = "=0.9.0"
```
nefrathenrici, line 15: Could you add `ClimaAtmos = "0.20.1"`?

experiments/surface_fluxes_perfect_model/Project.toml
```diff
@@ -0,0 +1,24 @@
...
+TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
+Thermodynamics = "b60c26fb-14c3-4610-9d3e-2d17fe7ff00c"
+YAML = "ddb6d928-2868-570f-bddf-ab3f9cf99eb6"
+
+[compat]
+Distributions = "0.25"
+EnsembleKalmanProcesses = "1"
+JLD2 = "0.4"
+TOML = "1"
+YAML = "0.4"
```
nefrathenrici, line 24: Could you add a SurfaceFluxes compat? Maybe 0.11 with Akshay's changes?

nefrathenrici (approved):
This looks great, thank you! I just had minor comments about the compat

experiments/pipeline.jl
```diff
@@ -0,0 +1,43 @@
...
+
+using Pkg
+experiment_id = ARGS[1]
+Pkg.activate("experiments/$experiment_id")
+Pkg.instantiate()
+using CalibrateAtmos
+pkg_dir = pkgdir(CalibrateAtmos)
+experiment_path = "$pkg_dir/experiments/$experiment_id"
+include("$experiment_path/model_interface.jl")
+# include("$experiment_path/generate_truth.jl")
```
akshaysridhar, line 11: Is this the synthetic data for the perfect-model problem ? Should this be uncommented for general use ? (I suppose if such a dataset exists then we don't need to regenerate it).

experiments/surface_fluxes_perfect_model/generate_truth.jl
```diff
@@ -0,0 +1,131 @@
...
+
+Generate synthetic observed y from the model truth.
+"""
+function synthetic_observed_y(x_inputs; data_path = "data", apply_noise = false)
+    config = YAML.load_file("$experiment_path/model_config.yml")
+    config["output_dir"] = data_path
+    y = obtain_ustar(FT, x_inputs, config, return_ustar = true)
+    if apply_noise
+        # add noise to model truth to generate synthetic observations with some observation error
+        Γ = 0.003^2 * I * (maximum(y) - minimum(y))
```
akshaysridhar, line 91:
```suggestion
        Γ = FT(0.003)^2 * I * (maximum(y) - minimum(y))
```

```
julia> Float64(0.003) * I * Float32(2)
UniformScaling{Float64}
0.006*I

julia> Float32(0.003) * I * Float32(2)
UniformScaling{Float32}
0.006*I
```

Does this matter?

akshaysridhar (approved):
Thanks @LenkaNovak, I've left brief comments. LGTM otherwise!

experiments/pipeline.jl
LenkaNovak: Oops yes, that's from when I was testing the atmos case. Good catch!

experiments/surface_fluxes_perfect_model/generate_truth.jl
LenkaNovak: Well, all the observables are converted to FT64 anyway (required by the EKI struct) but it's cleaner to include this explicitly, as you suggest.

PR #66, opened 2024-03-12
Add surfacefluxes test to CI

nefrathenrici:
This PR adds an end-to-end test for the surfacefluxes perfect model experiment.

Content
- Add `calibrate`, which runs the full calibration pipeline in a single function. This requires that the proper environment and model interface has already been loaded
- Use `calibrate` in the pipeline.jl (tested locally)
- Add end-to-end test using `calibrate` in `test/test_sf.jl`. I am a bit concerned that the parameter values are incorrect, since they do not match the default parameter values. 
- Add a unit test to ensure `calibrate` errors properly without the correct model interface
- resolve the surfacefluxes experiment Manifest
- Update the GHA

### DIFF
--- a/experiments/pipeline.jl
+++ b/experiments/pipeline.jl
@@ -9,35 +9,7 @@ pkg_dir = pkgdir(CalibrateAtmos)
 experiment_path = "$pkg_dir/experiments/$experiment_id"
 include("$experiment_path/model_interface.jl")
 include("$experiment_path/generate_truth.jl")
-ekp_config =
-    YAML.load_file(joinpath("experiments", experiment_id, "ekp_config.yml"))
 
-# initialize the CalibrateAtmos
-CalibrateAtmos.initialize(experiment_id)
-
-# run experiment with CalibrateAtmos for N_iter iterations
-N_iter = ekp_config["n_iterations"]
-N_mem = ekp_config["ensemble_size"]
-for i in 0:(N_iter - 1)
-    # run G model to produce output from N_mem ensemble members
-    physical_model =
-        CalibrateAtmos.get_forward_model(Val(Symbol(experiment_id)))
-
-    for m in 1:N_mem # TODO: parallelize with threads!
-        # model run for each ensemble member
-        model_config =
-            CalibrateAtmos.get_config(physical_model, m, i, experiment_id)
-        CalibrateAtmos.run_forward_model(physical_model, model_config)
-        @info "Finished model run for member $m at iteration $i"
-    end
-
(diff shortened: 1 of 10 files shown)


test/test_emulate_sample.jl
```diff
@@ -1,34 +0,0 @@
-import JLD2
-import Statistics: mean
```
LenkaNovak, line 2: why is this test deleted?

experiments/pipeline.jl
```diff
@@ -4,40 +4,12 @@ using Pkg
 experiment_id = ARGS[1]
 Pkg.activate("experiments/$experiment_id")
 Pkg.instantiate()
-using CalibrateAtmos
+import CalibrateAtmos
```
LenkaNovak, line 7: Just curious why we need import here, since we're importing in the included files.

LenkaNovak (approved):
LGTM, @nefrathenrici ! Looks much cleaner. I only had some clarification comments.

src/ekp_interface.jl
```diff
@@ -131,4 +131,57 @@ function update_ensemble(
...
+
+    # run experiment with CalibrateAtmos for N_iter iterations
+    N_iter = ekp_config["n_iterations"]
+    N_mem = ekp_config["ensemble_size"]
+    eki = nothing
+    for i in 0:(N_iter - 1)
+        # run G model to produce output from N_mem ensemble members
+        physical_model = get_forward_model(Val(Symbol(experiment_id)))
+
+        for m in 1:N_mem # TODO: parallelize with pmap!
```
LenkaNovak, line 169: Is is worth adding the pmap here or shall we not bother for now?

test/test_emulate_sample.jl
nefrathenrici: I thought that `emulate_sample` is a bit misleading if I add a test for the `calibrate` function. However, that test should really go in the model interface tests, and this can stay `emulate_sample`!

experiments/pipeline.jl
nefrathenrici: I don't think we need import here, thank you for catching that!

src/ekp_interface.jl
nefrathenrici: I tried to add ClimaComms multithreading quickly, but I ran into issues with multiple threads reading from JLD2, so I think I will leave it for another PR.

nefrathenrici: Actually, I went back and added locking around the JLD2 reading, so we now have basic multithreading!

PR #87, opened 2024-05-21
Add ClimaGPU Backend

nefrathenrici:
## Content

----

### DIFF
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -1,20 +1,42 @@
 abstract type AbstractBackend end
 
 struct JuliaBackend <: AbstractBackend end
-struct CaltechHPC <: AbstractBackend end
+abstract type SlurmBackend <: AbstractBackend end
+struct CaltechHPCBackend <: SlurmBackend end
+struct ClimaGPUBackend <: SlurmBackend end
 
 """
     get_backend()
 
-Determine the appropriate backend using relevant system information.
+Get ideal backend for deploying forward model runs. 
+Each backend is found via `gethostname()`. Defaults to JuliaBackend if none is found.
 """
 function get_backend()
-    backend = JuliaBackend
-    if isfile("/etc/redhat-release") &&
-       occursin("Red Hat", read("/etc/redhat-release", String))
-        backend = CaltechHPC
+    hostname = gethostname()
+    if occursin(r"^hpc-(\d\d)-(\d\d).cm.cluster$", hostname) ||
+       occursin(r"^login[1-4].cm.cluster$match", hostname)
+        return CaltechHPCBackend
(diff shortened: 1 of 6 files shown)


src/backends.jl
```diff
@@ -1,17 +1,50 @@
...
+    get_backend()
+
+Get ideal backend for deploying forward model runs.
+
+Each backend is found via specific system information:
+- JuliaBackend: default
+- CaltechHPC: String "Red Hat" in /etc/redhat-release
+- ClimaGPU: Correct GPU names returned from `nvidia-smi --query-gpu=gpu_name --format=csv``
+"""
 function get_backend()
```
Sbozzolo, line 18: A better way to identify a backend is using the `hostname`.
`hostname` for `clima` is `clima.gps.caltech.edu`, for `CaltechHPC`, the hostname matches the regular expression `$login[1-4].cm.cluster^`.

src/backends.jl
```diff
@@ -1,20 +1,42 @@
...
+Get ideal backend for deploying forward model runs. 
+Each backend is found via `gethostname()`. Defaults to JuliaBackend if none is found.
 """
 function get_backend()
-    backend = JuliaBackend
-    if isfile("/etc/redhat-release") &&
-       occursin("Red Hat", read("/etc/redhat-release", String))
-        backend = CaltechHPC
+    hostname = gethostname()
+    if occursin(r"^hpc-(\d\d)-(\d\d).cm.cluster$", hostname) ||
```
Sbozzolo, line 16: I think `match` is more appropriate function to use here

nefrathenrici: Is `!isnothing(match(regex, hostname))` then preferable to `occursin(regex, hostname)`, or is there a better way to use `match`?

Sbozzolo: One might wonder why you used `occursin`. What you want to check is that the hostname matches your regular expression, non that it contains it. 

But what you could do is (schematically):
```julia
HOSTNAMES = Dict("^clima.gps.caltech.edu$" => ClimaGPU, "aaaa" => CaltechHPC, "bbb" => CaltechHPC)
backend = get(HOSTNAMES, first(filter(x -> !isnothing(match(x, gethostname(), keys(HOSTNAMES))))
```
(There's certainly more elegant ways)

nefrathenrici: I believe `occursin(regex,str)` checks that the string matches the regex, but your way is already cleaner. I will incorporate it soon, thank you!

Sbozzolo: Yes, it does, but when it is less clear to readers (I think). Your intent is to match the regular expression and you can make it more explicit by using the function match

PR #131, opened 2025-01-24
Fix srun argument passing for SlurmManager

nefrathenrici:
## Content

----


nefrathenrici:
The tests [pass](https://buildkite.com/clima/climacalibrate-clima-gpu-server/builds/278) on the clima GPU server but are timing out because Caltech central HPC is so busy. I am going to bypass the rules and merge.

PR #149, opened 2025-03-13
Add support for asynchronous workers

nefrathenrici:
## Purpose 

This PR enables workers to be added while a calibration is already underway.

## To-do

- [ ] Add `redirect_stderr(stdout)` to worker logger func

## Content

- Add an `expr` field to PBSManager and SlurmManager. This `expr` is evaluated just after the worker is initialized. This can be set via `SlurmManager(nprocs; expr)`. This enables workers to be initialized and load code asynchronously, without relying on `@everywhere`.
- Add method specialization for `Distributed.create_worker(manager::Union{SlurmManager, PBSManager},...)` in order to ensure the expression gets evaluated before any other code.
- Changed `run_worker_iteration` to check for newly added workers for each forward model run. The function now gets initial workers, then makes a vector of work (fwd model calls). While this vector is not empty, we check for new workers, and if workers are available we assign one unit of work to the worker. If no workers are available, wait for workers

Previous workflow:
```julia
addprocs(SlurmManager(nprocs)...)
@everywhere
    include(model_interface_file)
end

calibrate(WorkerBackend, ...)
```
New workflow:
```julia
expr = quote
    include(model_interface_file)
end
@async addprocs(SlurmManager(nprocs; expr)...)

calibrate(WorkerBackend, ...)
```

----

### DIFF
--- a/src/workers.jl
+++ b/src/workers.jl
@@ -52,31 +52,38 @@ end
 worker_cookie_arg() = `--worker=$(worker_cookie())`
 
 """
-    SlurmManager(ntasks=get(ENV, "SLURM_NTASKS", 1))
+    SlurmManager(; ntasks=get(ENV, "SLURM_NTASKS", 1), expr=:())
 
-The ClusterManager for Slurm clusters, taking in the number of tasks to request with `srun`.
+A ClusterManager implementation for Slurm job scheduling systems. Workers inherit the current Julia environment by default.
 
-To execute the `srun` command, run `addprocs(SlurmManager(ntasks))`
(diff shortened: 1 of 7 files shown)


src/workers.jl
```diff
@@ -406,6 +456,7 @@ This function should be called from the worker process.
 """
 function set_worker_logger()
     @eval Main using Logging
+    redirect_stderr(stdout)
```
nefrathenrici, line 459: This is incorrect and should be `redirect_stderr(io)`

PR #163, opened 2025-04-09
Add post_observation_map_hook and postprocess_g_ensemble

### LINKED ISSUE #161: Additional output ensemble analysis stage within each iteration
## Issue

Our current pipeline wraps the parallel call to the forward map (and the EKP update) within the following function
https://github.com/CliMA/ClimaCalibrate.jl/blob/daaedfbaaa12f7fbfa30f56390ef1e42d9f0b024/src/ekp_interface.jl#L378-L383

Often it would be useful to provide some brief analysis of the ensemble, for example user defined plots at this stage.

Why is this cumbersome right now? Mainly because functions either are defined over the ensemble member, else they do not have access to any of the `ekp` information. So ideally I think we want a function that operates on `g_ensemble` along with `iteration,ekp,prior,...` etc. with the purpose of plotting. 

Below I add a selection that could also allow modification of `g_ensemble` with information that it cannot access in `observation_map(iteration)` easily - though I guess this is a different question.

## Possible solution

Allow users to define a method `postprocess_and_analyse_g_ensemble` that can
- edit `g_ensemble` before the ekp update ("postprocess")
- plot and save quantities of interest from the ensemble ("analyse")

```julia
# default behaviour
function postprocess_and_analyse_g_ensemble(ekp, g_ensemble, output_dir, iteration, prior) 
    # some default plot?
    return g_ensemble
end

function observation_map_and_update!(ekp, output_dir, iteration, prior)
    g_ensemble = observation_map(iteration)
    save_G_ensemble(output_dir, iteration, g_ensemble)
    processed_g_ensemble = analyze_g_ensemble(ekp, g_ensemble, output_dir, iteration, prior)
    terminate = update_ensemble!(ekp, processed_g_ensemble, output_dir, iteration, prior)
    return terminate
end
```

ph-kev:
closes #161 - This PR adds `post_observation_map_hook` and `postprocess_g_ensemble`.

## TODO

- [x] Add documentation about these functions

### DIFF
--- a/src/model_interface.jl
+++ b/src/model_interface.jl
@@ -1,7 +1,7 @@
 import EnsembleKalmanProcesses as EKP
 import YAML
 
-export forward_model, observation_map
+export forward_model, observation_map, analyze_iteration
 
 """
     forward_model(iteration, member)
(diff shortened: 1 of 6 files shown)


src/model_interface.jl
```diff
@@ -24,3 +25,34 @@ This function must be implemented for each calibration experiment.
...
+`g_ensemble`.
+"""
+function post_observation_map_hook(
+    ekp,
+    g_ensemble,
+    prior,
+    output_dir,
+    iteration,
+)
+    return nothing
```
nefrathenrici, line 47: Could you make this log the mean constrained parameters and the error for the previous iteration by default?

nefrathenrici (approved):
Thank you! One last thing - in the [buildkite test case](https://buildkite.com/clima/climacalibrate-derecho/builds/160#01963bb4-4265-4f70-aa65-9a52f66b22e2/712-761), the default `analyze_iteration` throws an error on the first iteration. I think this is because `analyze_iteration` needs to come after `update_ensemble!` [here](https://github.com/CliMA/ClimaCalibrate.jl/pull/163/files#diff-cb14ab4c2e5e66bf900b7c47dd9676bb6efd88940c2c9bbaffcd0371acbbfc5dR381-R384).

PR #165, opened 2025-04-14
Fix failure_rate kwarg in WorkerBackend, improve error message

### LINKED ISSUE #164: error when exceeding model failure rate should inform user that this is why it fails
[Here](https://github.com/CliMA/ClimaCalibrate.jl/blob/e040d672e80faacb916bb9cb520d77f6481b7be8/src/workers.jl#L43) the worker's code throws an error if the failure rate exceeds some default threshold rate but doesn't indicate that the failure rate is the source of the error. The message should be improved to indicate this. 

For example we have:
```out
ERROR: LoadError: Ensemble for iter 0 had a 56.00000000000001% failure rate
```
It would be more informative to have a message like 
```bash
ERROR: ErrorException: Halting execution because iter 0 had a 56% failure rate which is higher than the DEFAULT_FAILURE_RATE.
```
The docstring for the calibrate function should also be updated to note this, e.g. [here](https://github.com/CliMA/ClimaCalibrate.jl/blob/e040d672e80faacb916bb9cb520d77f6481b7be8/src/backends.jl#L233)

nefrathenrici:
## Purpose 

Closes #164 

## Content

Fixes the failure_rate kwarg when using the WorkerBackend and improved the error message to be more informative

----

### DIFF
--- a/src/workers.jl
+++ b/src/workers.jl
@@ -17,8 +17,6 @@ function run_worker_iteration(
     worker_pool = default_worker_pool(),
     failure_rate = DEFAULT_FAILURE_RATE,
 )
-    # Create a channel to collect results
-    results = Channel{Any}(ensemble_size)
     nfailures = 0
     @sync begin
         for m in 1:(ensemble_size)
@@ -39,8 +37,10 @@ function run_worker_iteration(
     end
     iter_failure_rate = nfailures / ensemble_size
     if iter_failure_rate > failure_rate
-        error(
-            "Ensemble for iter $iter had a $(iter_failure_rate * 100)% failure rate",
+        throw(
+            ErrorException(
+                "Execution halted: Iteration $iter had a $(round(iter_failure_rate * 100; digits=2))% failure rate, exceeding the maximum allowed threshold of $(failure_rate * 100)%.",
+            ),
         )
     end
 end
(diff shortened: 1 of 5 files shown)


src/workers.jl
```diff
@@ -39,8 +39,10 @@ function run_worker_iteration(
     end
     iter_failure_rate = nfailures / ensemble_size
     if iter_failure_rate > failure_rate
-        error(
-            "Ensemble for iter $iter had a $(iter_failure_rate * 100)% failure rate",
+        throw(
+            ErrorException(
+                "Execution halted: Iteration 0 had a $(round(iter_failure_rate * 100; digits=2))% failure rate, exceeding the maximum allowed threshold of $(failure_rate * 100)%.",
```
ph-kev, line 44:
```suggestion
        error(
                "Execution halted: Iteration $iter had a $(round(iter_failure_rate * 100; digits=2))% failure rate, exceeding the maximum allowed threshold of $(failure_rate * 100)%.",
```

ph-kev (changes requested):

src/workers.jl
nefrathenrici: Ah, that is a sloppy typo. Thanks for catching it!

ph-kev (approved):

Julians42 (approved):
Looks good - thanks Nat!

PR #166, opened 2025-04-21
Add forward model checkpointing

### LINKED ISSUE #158: Add support for forward model restarts
At the start of the forward model, we should indicate that the model is in progress, and update this indicator once the model has finished (regardless of error). This could easily be done by saving a file in each member's directory and updating it.

### LINKED ISSUE #160: Documentation lacking for restarts
There is no documentation for how restarting a calibration work. I assume that it is implemented from this [PR](https://github.com/CliMA/ClimaCalibrate.jl/pull/141).

nefrathenrici:
## Purpose 

This PR adds restarts/checkpointing for forward model runs. Closes #158, closes #160

## Content

- Added forward model checkpointing via simple file checks in each member's folder. This is not supported on the JuliaBackend.
- Split the inner iteration loop in the `calibrate` method for `HPCBackend` into its own function to use in testing.
- Added tests to ensure proper results with partial iterations.
- Added documentation for model and iteration-level checkpointing.

----


src/workers.jl
```diff
@@ -18,21 +18,38 @@ function run_worker_iteration(
...
-                    put!(worker_pool, worker)
-                end
+
+    work_to_do = map(1:ensemble_size) do m
+        (w) -> begin
+            if model_completed(output_dir, iter, m)
+                @info "Skipping completed member $m (found checkpoint)"
+                return
+            elseif model_started(output_dir, iter, m)
+                @info "Restarting member $m on worker $w (incomplete run detected)"
```
ph-kev, line 28: I might just be confused, but does this refer to restarting because the checkpoint is not found and so it is restarting from the beginning or restarting from a checkpoint?

ph-kev (approved):
Could we add some documentation somewhere that explain how model restarts work with respect to calibration? For example, what should the user expect if there is a checkpoint to restart from and there is no checkpoint to restart from?

src/workers.jl
nefrathenrici: This refers to restarting because the checkpoint is found but not marked as complete. Perhaps clearer wording would be resuming rather than restarting. We cannot provide a generic way to restart a model from a checkpoint, so the two else/elseif cases are functionally the same if the model does not have a way to restart from a checkpoint.

PR #172, opened 2025-04-25
Pretty format iteration time

### LINKED ISSUE #171: Pretty format iteration time
As of now, the iteration time is just a number without any units attached to indicate that they're seconds. It would be helpful to round to the nearest second and format it, so it is easier to read.

ph-kev:
closes #171 - This PR adds Dates as a dependency and pretty format the iteration time.


nefrathenrici (approved):
Thanks! It looks like the tests do not have the packages properly resolved. I think you need to add `Pkg.resolve()` to the buildkite pipelines, that probably should have been there before.
This should be sufficient:
```
      - julia --project=experiments/surface_fluxes_perfect_model -e 'using Pkg; Pkg.resolve(); Pkg.instantiate(;verbose=true)'
```

PR #173, opened 2025-04-25
Remove unused ExperimentConfig struct

nefrathenrici:
## Purpose 

This PR simplifies the source code by removing the unused ExperimentConfig struct. We can always add a new struct if we decide we need an extra layer to store experiment information.

## Content

- Remove `ExperimentConfig`
- Simplify backend dispatch to filter kwargs instead of having a method per backend. This will keep the code more maintainable as we add new backends.
- Updated surface fluxes test compat and Manifest

----

### DIFF
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -137,26 +137,6 @@ function calibrate(
 end
 
 function calibrate(
-    b::Type{JuliaBackend},
-    ensemble_size,
-    n_iterations,
-    observations,
-    noise,
-    prior,
-    output_dir;
-    ekp_kwargs...,
-)
-    ekp = ekp_constructor(
-        ensemble_size,
-        prior,
-        observations,
-        noise;
-        ekp_kwargs...,
-    )
-    return calibrate(b, ekp, n_iterations, prior, output_dir)
-end
-
-function calibrate(
     ::Type{JuliaBackend},
     ekp::EKP.EnsembleKalmanProcess,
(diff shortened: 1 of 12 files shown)


src/backends.jl
```diff
@@ -256,56 +198,54 @@ WorkerBackend uses Distributed.jl to run the forward model on workers.
...
+    ),
+)
+
+# Filter `calibrate` kwargs for the given backend.
+# Removes unused kwargs and merges result with defaults.
+function filter_calibrate_kwargs(b::Type{<:AbstractBackend}, kwargs)
+    default_kws = default_kwargs(b)
+    kwarg_keys = keys(default_kws)
+    filtered_kwargs = filter(x -> first(x) in kwarg_keys, pairs(kwargs))
+    return merge(default_kws, filtered_kwargs)
```
ph-kev, line 248: I am not sure if it is helpful here, but you can pass in keyword arguments like this: `f(; x = 1, y = 2, (y = 100,)...)`, where `y = 100` will override `y = 2`. Although, you can't do `f(; x = 1, y = 2, y = 100)`.

ph-kev (approved):

src/backends.jl
nefrathenrici: I think that the `merge` is a bit clearer than splatting out the overriding kwargs, but I didn't think of that, thanks.

PR #191, opened 2025-06-05
Add ObservationRecipe

ph-kev:
This PR adds ObservationRecipe which gives a way to construct observations with seasonal noise covariance matrix or SVDplusD covariance matrix.

# TODO

- [ ] Make a new release of ClimaAnalysis
- [ ] Add skipping of tests depending on EKP version since tests currently rely on the main branch 
- [ ] Check documentation and tests passed with new version of ClimaAnalysis


ext/ClimaAnalysisExt.jl
```diff
@@ -0,0 +1,350 @@
...
+                start_date,
+                end_date;
+                name = nothing)
+
+Return an `EKP.Observation` with a sample between the dates `start_date` and
+`end_date`, a covariance matrix defined by `covar_estimator`, `name` determined
+from the short names of `vars`, and metadata.
+
+!!! note
+    Metadata is only added with EnsembleKalmanProcesses v2.4.3 or later.
```
nefrathenrici, line 209:
```suggestion
    Metadata is only added with EnsembleKalmanProcesses v2.4.2 or later.
```

docs/src/observation_recipe.md
```diff
@@ -0,0 +1,100 @@
+# ObservationRecipe
+
+!!! warning
+    To enable this extension, use `using ClimaAnalysis` or `export
```
nefrathenrici, line 4:
```suggestion
    To enable this extension, use `using ClimaAnalysis` or `import
```

docs/src/observation_recipe.md
```diff
@@ -0,0 +1,100 @@
...
+After preprocessing the `OutputVar`s, so that they represent time series data of
+summary statistics, one can use set up an `EKP.observation` as shown below.
+
+```julia
+import ClimaAnalysis
+import EnsembleKalmanProcesses as EKP
+import ClimaCalibrate
+import ClimaCalibrate.ObservationRecipe
+
+# vars are OutputVars preprocessed with ensure consistency with units, times,
```
nefrathenrici, line 51:
```suggestion
# vars are OutputVars preprocessed to ensure consistent units, times,
```

docs/src/observation_recipe.md
```diff
@@ -0,0 +1,100 @@
...
+# ClimaAnalysis.average_season_across_time
+vars = ClimaAnalysis.average_season_across_time.(vars)
+
+# We choose SVDplusDCovariance. We need to supply the start and end dates of
+# the samples with `sample_dates`. It is helpful to use
+# `ClimaAnalysis.dates(var)` to find the correct dates.
+sample_dates = [
+        (Dates.DateTime(i, 12, 1), Dates.DateTime(i + 1, 9, 1)) for
+        i in 2007:2009
+    ]
```
nefrathenrici, line 63: Maybe this could be a helper function since I expect some form of this will be repeated in calibration pipelines.

src/observation_recipe.jl
```diff
@@ -0,0 +1,167 @@
...
+- `sample_dates`: The start and end dates of each samples. This is used to
+determine the sample from the time series data of the `OutputVar`s. These dates
+must be present in all the `OutputVar`s.
+
+Keyword arguments
+=====================
+
+- `model_error_scale`: Noise from the model error added to the covariance
+matrix. This is `(model_error_scale * mean(samples, dims = 2)).^2`, where
+`mean(samples, dims = 2)` is the mean of the samples.
```
nefrathenrici, line 128: Could you add a note surfacing the issue of not using years of data with `model_error_scale` ? You left a nice comment [here](https://github.com/CliMA/ClimaCalibrate.jl/pull/191/files#diff-c3d90bfa0e6a7f30b9ed65437de69858dcbe33ffb796b57f0abe308d963d1a59R178-R182) but I think it would be good to inform users as well.

nefrathenrici (commented):
Thanks, the code is structured cleanly. I only really had documentation comments.

ext/ClimaAnalysisExt.jl
ph-kev: For this, I was talking about embedding metadata with observation which is currently only available in the main branch, but not available in v2.4.2, rather than embedding metadata in an observation series. It should probably be rewritten to be a version later than v2.4.2, since v2.4.3 doesn't exist.

docs/src/observation_recipe.md
```diff
@@ -0,0 +1,100 @@
...
+
+## How do I use this to set up observation for calibration with EKP?
+
+All functions assumes that any data preprocessing is done with `ClimaAnalysis`.
+
+### Covariance Estimators
+
+There are currently two covariance estimators which are
+`SeasonalDiagonalCovariance` and `SVDplusDCovariance` which are subtypes of
+`AbstractCovarianceEstimator`.
```
Julians42, line 20:
```suggestion
There are currently two covariance estimators,
`SeasonalDiagonalCovariance` and `SVDplusDCovariance`, which are subtypes of
`AbstractCovarianceEstimator`.
```

docs/src/observation_recipe.md
```diff
@@ -0,0 +1,100 @@
...
+
+The `OutputVar`s should represent time series data of summary statistics. For
+example, to compute seasonal averages of a `OutputVar`, one can use
+`ClimaAnalysis.average_season_across_time`, which will produce a `OutputVar`
+that can be used with either `SeasonalDiagonalCovariance` or
+`SVDplusDCovariance`.
+
+```julia
+import ClimaAnalysis
+ClimaAnalysis.average_season_across_time(var)
```
Julians42, line 37: Maybe would be more MWE if you had a line reading temp var, for example from a simulation file?

docs/src/observation_recipe.md
```diff
@@ -0,0 +1,100 @@
...
+end_date = sample_dates[1][2]
+obs = ObservationRecipe.observation(covar_estimator, vars, start_date = start_date, end_date = end_date)
+```
+
+## Common questions
+
+**Q: I need to compute `g_ensemble` and I do not know how the data of the `OutputVar`s are flattened.**
+
+**A:** When forming the sample, the data in a `OutputVar` is flattened using
+`ClimaAnalysis.flatten`. See the ClimaAnalysis documentation for
```
Julians42, line 81: great that we have a consistent way to flatten observations!

ext/ClimaAnalysisExt.jl
nefrathenrici: Sounds good! Maybe we can bug Ollie for a release.

ext/ClimaAnalysisExt.jl
```diff
@@ -0,0 +1,350 @@
...
+                    seasonal_average_var.data
+                ) .^ 2,
+            )
+        end
+
+        # Construct full covariance matrix
+        # First, find the seasons covered by the dates
+        seasons = find_seasons(start_date, end_date)
+        seasons_to_int =
+            Dict("MAM" => 1, "JJA" => 2, "SON" => 3, "DJF" => 4)
```
Julians42, line 105: Do we need this hard-coded mapping pattern here?

docs/src/observation_recipe.md
Julians42: Hmm yeah points is maybe not the right word. How about "different observations"?

src/observation_recipe.jl
```diff
@@ -0,0 +1,167 @@
...
+    )
+end
+
+"""
+    SVDplusDCovariance <: AbstractCovarianceEstimator
+
+Contain the necessary information to construct a `EKP.SVDplusD` covariance
+matrix from `ClimaAnalysis.OutputVar`s.
+"""
+struct SVDplusDCovariance{ # TODO: Should change the name of this to `SVDplusD`
```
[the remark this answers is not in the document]
ph-kev: This is a poorly worded TODO. It should be renaming that struct to `SVDplusDCovariance` instead of `SVDPlusDCovariance`. I think I forgot to remove that TODO. 

I also don't think there will be any name collision because in EKP, it is called `SVDplusD`.

ext/ClimaAnalysisExt.jl
ph-kev: This was here, because we know the mapping between season to index along the time dimension. This comment did help me realize there is a case that isn't covered.

ext/ClimaAnalysisExt.jl
```diff
@@ -0,0 +1,350 @@
...
+        error("$start_date should not be later than $end_date")
+
+    vars = _vars_to_iterable(vars)
+
+    # Check if all dates in vars are valid
+    _check_dates_in_var(vars, start_date, end_date)
+
+    # Get the flattened sample and metadata
+    windowed_vars =
+        ClimaAnalysis.window.(vars, "time", left = start_date, right = end_date)
```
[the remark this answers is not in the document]
ph-kev: I am not sure what this comment is referring to. Here, we are taking only the data between `start_date` and `end_date`.

If you are referring to the averaging that is done in `SeasonalDiagonalCovariance` when computing the model error scale, the time average is computed over the same season and that is added to the diagonal as the model error scale (with scaling of course).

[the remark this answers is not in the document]
Julians42: Ah yes I see now! You're right but I missed this and I did read the docs page ;), I left a comment there on my suggested change but yes maybe add a note to the doc string as well for this.

docs/src/observation_recipe.md
```diff
@@ -0,0 +1,100 @@
...
+There are currently two covariance estimators which are
+`SeasonalDiagonalCovariance` and `SVDplusDCovariance` which are subtypes of
+`AbstractCovarianceEstimator`.
+
+For `SeasonalDiagonalCovariance`, the time series data of seasonal quantities
+should be used, while for `SVDplusDCovariance`, any time series data of any
+summary statistics can be used. Look at the documentation of these functions for
+more information.
+
+### Recommendations for data preprocessing
```
Julians42, line 27: Based on discussion below, we should change this from `Recommendations` to something stronger - maybe an important flag or something? or `Necessary data preprocessing` since this preprocessing is essential (not just recommended) for the observation series below to work, e.g., we have to have reduced this to the correct spatial-temporal resolution before passing to the observations function.

Julians42 (approved):
Yes this looks good to me. Thanks for addressing the comments.

nefrathenrici (approved):
Thanks! The documentation is great. My single nitpick would be to add tests for the observation names but I think that the test suite is large enough as is. 
Could you increment the patch number in a separate commit?

PR #193, opened 2025-06-05
Add missing entries to compat

ph-kev:
This commit adds an entry for ClimaParams and Conda and updates the Aqua test for checking compat dependencies.


nefrathenrici (approved):
Thanks!

PR #197, opened 2025-06-25
Add ObservationRecipe.reconstruct_g_mean_final

ph-kev:
This PR adds `ObservationRecipe.reconstruct_g_mean_final`.

**TODO**

- [x] Write test for `_get_metadata_from_nth_iteration`
- [x] Clean up how indices are computed
- [x] Add any assumptions that are needed for the functions to run
- [x] Split single commit into two separate commits
- [x] > EnsembleKalmanProcesses v2.4.2


ext/ClimaAnalysisExt.jl
```diff
@@ -484,4 +484,57 @@ function _lat_weights_var(var::OutputVar; min_cosd_lat = 0.1)
...
+end
+
+"""
+    get_metadata_from_nth_iteration(obs_series, N)
+
+For the `N`th iteration, get the metadata of the observation(s) being processed.
+"""
+function ObservationRecipe.get_metadata_from_nth_iteration(obs_series, N)
+    num_epoches = EKP.get_length_epoch(obs_series)
+    minibatch_indices = EKP.get_minibatch(obs_series, mod1(N, num_epoches))
```
ph-kev, line 534: I wasn't sure if there is a better way of doing this with EKP.

docs/src/observation_recipe.md
```diff
@@ -88,6 +88,28 @@ obs = ObservationRecipe.observation(
...
+can be accessed with `EKP.get_metadata(obs::EKP.Observation)` and the metadata
+for each iteration can be accessed with
+[`ObservationRecipe.get_metadata_for_nth_iteration`](@ref). The metadata can be
+used with `ClimaAnalysis.unflatten` to reconstruct the original `OutputVar`
+before flattening. See the ClimaAnalysis
+[documentation](https://clima.github.io/ClimaAnalysis.jl/dev/api/#FlatVar) about
+`ClimaAnalysis.FlatVar` for more information.
+
+`ObservationRecipe` provides a helper function for reconstructing the mean
+formward map evaluation with [`ObservationRecipe.reconstruct_g_mean_final`](@ref
```
nefrathenrici, line 110:
```suggestion
forward map evaluation with [`ObservationRecipe.reconstruct_g_mean_final`](@ref
```

ext/ClimaAnalysisExt.jl
```diff
@@ -484,4 +484,61 @@ function _lat_weights_var(var::OutputVar; min_cosd_lat = 0.1)
...
+"""
+function ObservationRecipe.reconstruct_g_mean_final(
+    ekp::EKP.EnsembleKalmanProcess,
+)
+    obs_series = EKP.get_observation_series(ekp)
+    metadatas = ObservationRecipe.get_metadata_for_nth_iteration(
+        obs_series,
+        EKP.get_N_iterations(ekp),
+    )
+    @assert eltype(metadatas) <: ClimaAnalysis.Var.Metadata
```
nefrathenrici, line 520: Nitpick, could you add a helpful error message to this assert?

ext/ClimaAnalysisExt.jl
```diff
@@ -484,4 +484,61 @@ function _lat_weights_var(var::OutputVar; min_cosd_lat = 0.1)
...
+"""
+    get_metadata_for_nth_iteration(obs_series, N)
+
+For the `N`th iteration, get the metadata of the observation(s) being processed.
+"""
+function ObservationRecipe.get_metadata_for_nth_iteration(obs_series, N)
+    num_epoches = EKP.get_length_epoch(obs_series)
+    # EKP.get_minibatch fails with N > num_epoches, so we use mod1 to go back to
+    # the first epoch which seems consistent with what EKP does
+    minibatch_indices = EKP.get_minibatch(obs_series, mod1(N, num_epoches))
```
nefrathenrici, line 496: I don't understand the mod1 behavior, is this wrapping back around on the indices once a full epoch has been completed?

ph-kev: The mod1 behavior is to just loop back once a full epoch is completed. I am not sure if this is behavior from EnsembleKalmanProcesses that we can depend on though. I just noticed that when all the observations are ran through, it would go back to the start.

nefrathenrici: I think this is correct - after an epoch, the full dataset should have been used so it should loop back for the next epoch.

PR #204, opened 2025-07-08
Fix latitude weights with NaN in data

### LINKED ISSUE #203: Adding latitude weights does not work with OutputVars with NaNs
There is an implicit assumption that the `OutputVar` does not contain `NaN`s. This is not always true (e..g with land simulations).

ph-kev:
closes #203 - There is a bug when using latitude weights with `NaN` in the data. This bug is resolved by inserting `NaN` in the latitude weights whether there is `NaN` in the data. The `NaN`s are removed when flattening the data.

### DIFF
--- a/ext/ClimaAnalysisExt.jl
+++ b/ext/ClimaAnalysisExt.jl
@@ -399,8 +399,8 @@ end
 """
     lat_weights_var(var::OutputVar)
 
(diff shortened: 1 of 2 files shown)


nefrathenrici (approved):
Thank you!

PR #217, opened 2025-08-07
Add default status if PBS job status fails

### LINKED ISSUE #213: Derecho Backend is broken
The Derecho backend is broken, caused by failing `qstat` commands. When we fix this we should add better error handling.

[Example failure](https://buildkite.com/clima/climacalibrate-derecho/builds/367#0198681b-c1ad-445b-9e32-6dc19e268fca)

nefrathenrici:
## Purpose 

Fixes #213 to make derecho job status polling more robust.

## Content

Updated the `job_status(jobid::PBSJobID)` function to return a default state `:RUNNING` if the `qstat` command fails silently. This uses a new function to get PBS job status `_qstat_output` which tries to obtain the job status in different formats and with a few retries.

----

### DIFF
--- a/src/pbs.jl
+++ b/src/pbs.jl
@@ -135,13 +135,24 @@ end
 
 Submit a job to the PBS Pro scheduler using qsub, removing unwanted environment variables.
 
-Unset variables: "PBS_MEM_PER_CPU", "PBS_MEM_PER_GPU", "PBS_MEM_PER_NODE"
+Unset variables: "PBS_MEM_PER_CPU", "PBS_MEM_PER_GPU", "PBS_MEM_PER_NODE", "PYTHONHOME", "PYTHONPATH", "PYTHONUSERBASE"
 """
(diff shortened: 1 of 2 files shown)


src/pbs.jl
```diff
@@ -135,13 +135,24 @@ end
...
+        "PYTHONHOME",
+        "PYTHONPATH",
+        "PYTHONUSERBASE",
+    )
     for k in unset_env_vars
         haskey(env, k) && delete!(env, k)
     end
+    # Disable user-site packages directory to prevent issues with Derecho's
+    # `qstat` python backend
+    env["PYTHONNOUSERSITE"] = "1"
```
ph-kev, line 155: Not sure if this exists, but is there documentation for what is set or not set as environment variables on Derecho? If there is, can you put a link to it somewhere in the documentation?

nefrathenrici: There is no documentation on Derecho but I can link to the actual `qstat` backend [repo](https://github.com/NCAR/qstat-cache)

PR #222, opened 2025-08-22
Add GEnsembleBuilder

### LINKED ISSUE #211: Add helper function to construct g ensemble
With metadata in the observation, there should be enough information to fully construct the G ensemble matrix. It should be expected that the user preprocesses the diagnostics produced from the simulation until the last step is to window and flatten the `OutputVar`.

To determine the times of the window, you can use the dates of the metadata. This assumes that the conventions around times are all the same between the observational and simulation data.

To determine the order of the observation, you can use the order of the short names in the metadata and the `EKP` object itself.

# Concerns

1. If there are two variables with the same short name (e.g., different reductions), then how should this be handled?

One solution is to make the user pass in something like `pr_avg` and `pr_max`. 

2. If more steps are needed after windowing and flattening, how should this be handled?

One solution is to give the user an out by exposing the lower level functions. For example, instead of construct the full G ensemble matrix, the lower level functions can partially build it. This allows the user to handle the special cases themselves. 

3. How should additional keyword arguments for `flatten` should be passed?

I am not too sure about this, since each variable might require different keyword arguments for `flatten`. In addition, you most likely want to use the same keyword arguments that you use when preprocessing the observational data.

### LINKED ISSUE #220: Typo in docs
There is """ in the middle of the documentation.

Image

### LINKED ISSUE #221: _data_size is replaced by _data_length
In the new version of ClimaAnalysis, `ClimaAnalysis.Var._data_size` is replaced by `ClimaAnalysis.Var._data_length`.

ph-kev:
closes #221, closes #211, closes #220

This PR adds `GEnsembleBuilder` to facilitate easy construction of the G ensemble matrix using the metadata in the observation.


ph-kev:
I think this is good enough for a review, but not ready to merge (still need to release a ClimaAnalysis version and missing/helpful features for this struct).

ext/ClimaAnalysisExt.jl
```diff
@@ -546,23 +547,55 @@ function ObservationRecipe.reconstruct_g_mean_final(
...
+    _get_minibatch_indices_for_nth_iteration(ekp::EKP.EnsembleKalmanProcess, N)
+
+Get the indices that correspond to each metadata for the minibatch of the `N`th
+iteration.
+
+Note that the `indices` field in the `EKP.observation` cannot be used as the
+multiple `OutputVar`s are flattened and concatenated together as a single
+vector.
+"""
+function _get_minibatch_indices_for_nth_iteration(obs_series, N)
```
nefrathenrici, line 576: It would be nice to expose this info to the user so they can see the mapping of indices to metadata/short names, it could be a helper function called something like `ranges_by_short_name`

ext/ensemble_builder.jl
```diff
@@ -0,0 +1,364 @@
...
+    g_ens::Matrix{FT}
+
+    """Dictionary which map short name to a vector of metadata associated with
+       the short name"""
+    metadata_by_short_name::Dict{String, Vector{METADATAINFO}}
+
+    """A bit matrix which keeps track of which entries are filled out in the
+    G ensemble matrix. The size of this matrix is the number of metadata for the
+    minibatch by the number of ensemble members"""
+    completed::BitMatrix
```
nefrathenrici, line 42: It would be nice to build on this and expose which vars (rather than indices) are missing from the g_ensemble matrix, something like `missing_short_names` or `missing_vars`.

ext/ensemble_builder.jl
```diff
@@ -0,0 +1,364 @@
...
+    GEnsembleBuilder{FT <: AbstractFloat}
+
+An object to help build G ensemble matrix by using the metadata stored in the
+`EKP.EnsembleKalmanProcess` object. Metadata must come from `ClimaAnalysis`.
+
+`GEnsembleBuilder` takes in preprocessed `OutputVar`s and automatically
+construct the corresponding G ensembe matrix for the current iteration of the
+calibration.
+"""
+struct GEnsembleBuilder{FT <: AbstractFloat, METADATAINFO <: MetadataInfo}
```
nefrathenrici, line 31: It would be nice to add a summary or show method for this struct just to show the size, short names of vars in one column and what percent are completed.

nefrathenrici (commented):
This looks great so far! I think we can push this a bit further and make more information easily available to the user.

ext/ensemble_builder.jl
```diff
@@ -0,0 +1,364 @@
...
+    order or there are more dates in `var` than the dates in `metadata`.
+    For example, if `var` contains monthly averages and `metadata` represents
+    seasonal averages, then this function will not return an error, because
+    all dates in `metadata` is in the set of all dates in `var`.
+"""
+function _match_dates(var::OutputVar, metadata::Metadata)
+    # Metadata comes from observational data and var is simulation data
+    # when calibrating
+    obs_dates = ClimaAnalysis.dates(metadata)
+    sim_dates = ClimaAnalysis.dates(var)
```
ph-kev, line 323: Add a check for unique dates

ph-kev:
The `_match_dates` function can be more general and be moved to ClimaAnalysis as `select` which is a function that allows for arbitrary indexing of dimensions by index or value. See this issue https://github.com/CliMA/ClimaAnalysis.jl/issues/323.

docs/make.jl
```diff
@@ -17,12 +17,9 @@ Literate.markdown(
     joinpath(@__DIR__, "src"),
 )
 
+ClimaAnalysisExt = Base.get_extension(ClimaCalibrate, :ClimaAnalysisExt)
 makedocs(
-    plugins = [bib],
```
nefrathenrici, line 21: Is there a reasons you removed the `plugins`?

ext/ClimaAnalysisExt.jl
```diff
@@ -274,6 +275,17 @@ function ObservationRecipe.observation(
...
     _check_dates_in_var(vars, start_date, end_date)
 
+    # Check if dates are unique and short name exists
+    for var in vars
+        allunique(ClimaAnalysis.dates(var)) || @warn(
+            "Dates in OutputVar with short name $(short_name(var)) are not unique. You will not be able to use GEnsembleBuilder",
+        )
+    end
+
+    any(==(""), ClimaAnalysis.short_name.(vars)) && @warn(
```
nefrathenrici, line 285: Should these just be warnings or actual errors?

ext/ClimaAnalysisExt.jl
nefrathenrici: Yes, it should take a short name and return the indices, it might be useful but I don't think it's necessary in this PR.

ext/ensemble_builder.jl
```diff
@@ -0,0 +1,512 @@
...
+end
+
+"""
+    GEnsembleBuilder(::Type{FT},
+                     ekp::EKP.EnsembleKalmanProcess)
+
+Construct a `GEnsembleBuilder` where the element type of the G ensemble matrix
+is `FT`.
+"""
+function EnsembleBuilder.GEnsembleBuilder(
```
nefrathenrici, line 55: This function could infer the FT from EKP instead of passing it as an argument.

ext/ensemble_builder.jl
```diff
@@ -0,0 +1,364 @@
...
+function _is_compatible_with_metadata(var::OutputVar, metadata::Metadata)
+    return _same_short_names(var, metadata) &&
+           _same_dim_names(var, metadata) &&
+           _same_dim_units(var, metadata) &&
+           _same_units(var, metadata) &&
+           # For the temporal dimension, only check if the times of metadata is
+           # a subset of the times of var, because _match_dates will get the
+           # correct dates for us
+           _compatible_dims_values(var, metadata)
+end
```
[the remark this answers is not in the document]
nefrathenrici: I agree, this will be an important addition in a future PR.

src/ekp_interface.jl
```diff
@@ -425,3 +425,18 @@ function observation_series_from_samples(
...
 end
+
+"""
+    g_ens_mat(eki::EKP.EnsembleKalmanProcess, FT)
+
+Construct an uninitialized G ensemble matrix of type `FT`.
+"""
+function g_ens_mat(
+    eki::EKP.EnsembleKalmanProcess,
+    ::Type{FT},
```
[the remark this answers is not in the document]
nefrathenrici: Following up, is this not possible?

docs/make.jl
ph-kev: This is a typo. It will be included back

PR #228, opened 2025-09-19
Refactor matching functionality to a separate module and add more functionality

ph-kev:
This PR refactors the matching of OutputVar to metadata to another module and add more functionality such as a verbose option and other `Matcher` that would be helpful.

TODO

- [x] Add documentation about checkers
- [x] Add an option to pass your own checkers when filling the G ensemble matrix

### DIFF
--- a/docs/src/ensemble_builder.md
+++ b/docs/src/ensemble_builder.md
@@ -68,6 +68,9 @@ In particular, the `OutputVar`s from the simulation data should match the
   metadata in the observations,
 - the units of the variables.
 
+For more information about the checks that are performed, see the
+[Checkers](#checkers) section.
+
 !!! info "Spinup and windowing times"
     Internally, the correct dates are matched between the observational and
     simulation data. As a result, you do not need to window the times (e.g. when
@@ -78,6 +81,13 @@ In particular, the `OutputVar`s from the simulation data should match the
     errors. For example, if the simulation data contain monthly averages and
     metadata track seasonal averages, then no error is thrown, because all dates
     in `metadata` are in all the dates in `var`.
+
+    For these reasons, it is recommended to pass
+    [`SequentialIndicesChecker`](@ref ClimaCalibrate.Checker.SequentialIndicesChecker)
+    to the `checkers` keyword argument of
+    [`fill_g_ens_col!`](@ref ClimaCalibrate.EnsembleBuilder.fill_g_ens_col!)
+    which will check that the dates used to fill the G ensemble matrix
+    correspond to sequential indices in the simulation data.
 
 After preprocessing the `OutputVar`s, you can call
 [`fill_g_ens_col!`](@ref ClimaCalibrate.EnsembleBuilder.fill_g_ens_col!) to fill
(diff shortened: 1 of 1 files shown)


ext/checkers.jl
```diff
@@ -0,0 +1,230 @@
...
+function Checker.check(
+    ::SequentialIndicesChecker,
+    var::OutputVar,
+    metadata::Metadata;
+    verbose = false,
+)
+    obs_dates = ClimaAnalysis.dates(metadata)
+    sim_dates = ClimaAnalysis.dates(var)
+    for dates in (obs_dates, sim_dates)
+        allunique(dates) || error("Dates in $dates are not unique")
```
nefrathenrici, line 216: Should this throw an error? I think this could just warn and return `false`.

nefrathenrici (approved):
Thank you! I like that this is easily extensible.

ext/ensemble_builder.jl
```diff
@@ -205,123 +243,27 @@ function _try_fill_g_ens_col_with_var!(
...
-    temporal_dim = try
-        ClimaAnalysis.dates(var_or_metadata)
-    catch
-        ClimaAnalysis.times(var_or_metadata)
-    end
-    return temporal_dim
+    return all(
+        Checker.check(checker, var, metadata; verbose = verbose) for
+        checker in checkers
+    )
```
nefrathenrici, line 266:
```suggestion
    for checker in checkers
        Checker.check(checker, var, metadata; verbose = verbose) || return false
    end
    return true
```

Minor, but this could exit early when there is a false check

src/checkers.jl
```diff
@@ -0,0 +1,89 @@
+module Checker
+
+"""
+    abstract type AbstractChecker end
+
+An object that performs validation checks between the simulational data and
```
nefrathenrici, line 6:
```suggestion
An object that performs validation checks between the simulation data and
```

ext/checkers.jl
```diff
@@ -0,0 +1,230 @@
...
+)
+    var_units = ClimaAnalysis.units(var)
+    metadata_units = ClimaAnalysis.units(metadata)
+    if var_units == "" || metadata_units == ""
+        @warn("Var or metadata may be missing units")
+    end
+    same_units = var_units == metadata_units
+    verbose &&
+        !same_units &&
+        @info "Units of OutputVar ($var_dim_name) are not the same as units of metadata ($var_dim_units)"
```
nefrathenrici, line 129:
```suggestion
@info "Units of OutputVar ($var_units) are not the same as units of metadata ($metadata_units)"
```

I think this is a typo

ext/ensemble_builder.jl
ph-kev: I think `all` does an early exit already.

```julia
julia> function foo(x)
       x == 1 && false
       x == 2 && error("foo")
       end
foo (generic function with 1 method)

julia> all(foo(x) for x in 1:2)
false

julia> all(foo(x) == false for x in 1:2)
ERROR: foo
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:35
 [2] foo
   @ ./REPL[14]:3 [inlined]
 [3] #9
   @ ./none:0 [inlined]
 [4] iterate
   @ ./generator.jl:48 [inlined]
 [5] _all
   @ ./reduce.jl:1316 [inlined]
 [6] all
   @ ./reduce.jl:1301 [inlined]
 [7] all(itr::Base.Generator{UnitRange{Int64}, var"#9#10"})
   @ Base ./reduce.jl:1193
 [8] top-level scope
   @ REPL[16]:1
 [9] top-level scope
   @ none:1
```

nefrathenrici: oh great, then this is unnecessary

ext/checkers.jl
ph-kev: I think it should throw an error, because there shouldn't be duplicate dates.

I can remove this though, since this is checked in `_match_dates`.

nefrathenrici (approved):

PR #239, opened 2025-10-07
Add run_initial_ensemble

### LINKED ISSUE #198: Add ability to run the first iteration without an EKP observation series
The motivation for this is that the observation in some sense depends on the model diagnostics. As of now, we need to run the simulation or know what the grid look like ahead of time to create the observation.

### LINKED ISSUE #202: Improvements to calibration pipeline
In addition to the issue for the rewrite of the land calibrate pipeline, there are improvements that can be made to all calibration pipelines.

1. **Error checking for units and grid.** There is no unit checking or checking if the grid are the same. This is can be fixed by using the metadata in the observation. Completed by https://github.com/CliMA/ClimaCalibrate.jl/pull/222 and https://github.com/CliMA/ClimaCalibrate.jl/pull/228
2. **Creating observation require knowing the grid, the order of variables in the observations, and potentially a mask.** To know the grid, the best way is to use the diagnostics. However, EKP requires building the observation first, so you can't use one of the ensemble members of the first iteration to create the observation. To know the order of the variables, you can use the short names of the metadata in the observation. Finally, a mask can be created from the diagnostics.

ph-kev:
closes #198, closes #202 - This PR adds `run_initial_ensemble` which allows one to run an ensemble without creating the entire `EKP.EnsembleKalmanProcess` object. This PR also add `run_forward_models` which is split off from the `calibrate` function as it needed to start the forward models.

TODO

- [ ] New release of EnsembleKalmanProcesses.jl

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -66,6 +66,7 @@ ClimaCalibrate.parameter_path
 ClimaCalibrate.minibatcher_over_samples
 ClimaCalibrate.observation_series_from_samples
 ClimaCalibrate.load_latest_ekp
+ClimaCalibrate.run_initial_ensemble
 ```
 
 ## Observation Recipe Interface
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -150,7 +150,7 @@ function calibrate(
 end
 
 function calibrate(
-    ::Type{JuliaBackend},
+    b::Type{JuliaBackend},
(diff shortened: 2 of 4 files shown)


ph-kev:
Closed because it is stale and the functionality should be implemented in EnsembleKalmanProcesses and not in ClimaCalibrate

PR #250, opened 2025-10-20
fix docstring of `analyze_iteration`

### LINKED ISSUE #249: `analyze_iteration` is run after `update_ensemble!`, but its docstring says otherwise
See this line:

https://github.com/CliMA/ClimaCalibrate.jl/blob/dc3845f8db8e67e2a30693f97d9550f5864da7d8/src/model_interface.jl#L31

which states that `analyze_iteration` is run _before_ `update_ensemble!`. However here:

https://github.com/CliMA/ClimaCalibrate.jl/blob/dc3845f8db8e67e2a30693f97d9550f5864da7d8/src/ekp_interface.jl#L346-L352

However in the source code `analyze_iteration` is run _after_ `update_ensemble!`. 

So either the source code or the docstring should be changed, but I think `analyze_iteration` should be run before `update_ensemble!`? This is so that diagnostics about the calibration can be made on the forward runs and the EKP object before it is updated. 

Or perhaps we need another helper that enables the user to also analyze the iteration after `update_ensemble!` to look at the updated EKP.

xkykai:
Closes #249

### DIFF
--- a/src/model_interface.jl
+++ b/src/model_interface.jl
@@ -28,7 +28,7 @@ end
 """
     analyze_iteration(ekp, g_ensemble, prior, output_dir, iteration)
 
-After each evaluation of the observation map and before updating the ensemble,
+After each evaluation of the observation map and the ensemble update,
 `analyze_iteration` is evaluated.
 
 This function is optional to implement.


ph-kev (approved):

ph-kev:
Closing in favor of https://github.com/CliMA/ClimaCalibrate.jl/pull/247

PR #256, opened 2025-10-23
Fix bug with SignChecker when NaNs are present

### LINKED ISSUE #254: Bug with SignChecker
The implementation of sign checker is to do the following computation

```julia
nanmean(var.data .> 0)
```

This doesn't work, because `NaN .> 0` is `false` and not `NaN`.

ph-kev:
The original code is `nanmean(var.data .> 0)`. This does not work, because `NaN > 0` is `false` and not `NaN`.

closes #254

### DIFF
--- a/ext/checkers.jl
+++ b/ext/checkers.jl
@@ -264,7 +264,8 @@ function Checker.check(
     # This is inaccurate, because not all the values in var.data will end up in
     # the G ensemble matrix. See _match_dates for one case. However, the mean
     # should not change that much with additional times.
-    sim_pos_proportion = nanmean(var.data .> 0)
+    valid = @. !isnan(var.data)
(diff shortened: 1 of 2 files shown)


ext/checkers.jl
```diff
@@ -264,7 +264,14 @@ function Checker.check(
...
     # should not change that much with additional times.
-    sim_pos_proportion = nanmean(var.data .> 0)
+    positive_count = 0
+    count = 0
+    for val in var.data
+        isnan(val) && continue
+        count += 1
+        positive_count += val > 0
+    end
+    sim_pos_proportion = positive_count / count
```
nefrathenrici: I think this could be simpler:
```suggestion
    valid = @. !isnan(var.data)
    sim_pos_proportion = sum((var.data .> 0) .& valid) / sum(valid)
```

PR #286, opened 2026-03-05
Add SVD residual analysis utilities

nefrathenrici:
## Purpose 

This PR adds tools to analyze model-data mismatch when using SVDPlusD noise covariances.

## Content

- `compute_structured_energy(diff, eigvectors)` scalar ratio of the residual's norm in the top-eigenvector subspace to the unstructured remainder. Values > 1 mean structured noise dominates.
- `compute_structured_energy_by_variable(diff, eigvectors, metadata_vec)` same ratio computed per variable
- `compute_normalized_projections(diff, eigvectors, eigvalues, metadata_vec)` returns a (n_eigenvectors x n_variables) matrix of z-scores `a_i / sqrt(lambda_i)`. Values >> 1 indicate mismatch beyond noise for that variable/eigenvector pair.
- `analyze_residual(ekp, iter; n_eigenvectors=3)` orchestrates the above, returning all four diagnostics (normalized_projections, structured_energy, structured_energy_by_variable, residual_norm_by_variable).
- Added ArnoldiMethod, LinearMaps as dependencies. These are small packages which don't introduce any large deps.

----

### DIFF
--- a/src/svd_analysis.jl
+++ b/src/svd_analysis.jl
@@ -1,0 +1,219 @@
+import LinearAlgebra: SVD, Diagonal, norm, dot, UniformScaling
+import LinearMaps: LinearMap
+import ArnoldiMethod: partialschur, partialeigen
+import Statistics: mean
+
+export compute_structured_energy,
+    compute_structured_energy_by_variable,
+    compute_normalized_projections,
+    analyze_residual
+
+# TODO: Use the function in EKP once it's moved there
+"""
+    _create_compact_linear_map(A)
+
+Wrap an SVD-structured covariance matrix (or a vector of them) as a `LinearMap`
+for use with matrix-free eigensolvers such as `ArnoldiMethod.partialschur`.
+
+Supports `LinearAlgebra.SVD`, `LinearAlgebra.Diagonal`, and `EKP.SVDplusD`
+inputs. If `A` is a vector, the matrices are treated as block-diagonal.
+"""
+function _create_compact_linear_map(A)
+    Avec = isa(A, AbstractVector) ? A : [A]
+
+    Us = []
+    Ss = []
+    VTs = []
+    ds = []
+    batches = []
+    shift = 0
+    for a in Avec
+        if isa(a, UniformScaling)
+            throw(
+                ArgumentError(
+                    "Detected `UniformScaling` (i.e. \"λI\") StructureMatrix, and unable to infer dimensionality. \n Please recast this as a diagonal matrix, defining \"λI(d)\" for dimension d",
+                ),
+            )
+        elseif isa(a, SVD)
+            push!(Us, a.U)
+            push!(Ss, a.S)
+            push!(VTs, a.Vt)
+            push!(ds, zeros(size(a.U, 1)))
+            bsize = size(a.U, 1)
+        elseif isa(a, Diagonal)
+            # No low-rank component; represent as empty U, S, Vt so the
+            # LinearMap formula U*(S.*(Vt*x)) + d.*x reduces to just d.*x
+            n = length(a.diag)
+            push!(Us, zeros(n, 0))
+            push!(Ss, zeros(0))
+            push!(VTs, zeros(0, n))
+            push!(ds, a.diag)
+            bsize = n
+        elseif isa(a, EKP.SVDplusD)
+            svda = a.svd_cov
+            diaga = (a.diag_cov).diag
+            push!(Us, svda.U)
+            push!(Ss, svda.S)
+            push!(VTs, svda.Vt)
+            push!(ds, diaga)
+            bsize = length(diaga)
+        else
+            throw(
+                ArgumentError(
+                    "Unsupported matrix type: $(typeof(a)). Expected SVD, Diagonal, or EKP.SVDplusD.",
+                ),
+            )
+        end
+
+        batch = (shift + 1):(shift + bsize)
+        push!(batches, batch)
+        shift = batch[end]
+    end
+
+    Amap = LinearMap(
+        x -> reduce(
+            vcat,
+            [
+                U * (S .* (Vt * x[batch])) + d .* x[batch] for
+                (U, S, Vt, d, batch) in zip(Us, Ss, VTs, ds, batches)
+            ],
+        ),
+        x -> reduce(
+            vcat,
+            [
+                Vt' * (S .* (U' * x[batch])) + d .* x[batch] for
+                (U, S, Vt, d, batch) in zip(Us, Ss, VTs, ds, batches)
+            ],
+        ),
+        sum(size(U, 1) for U in Us),
+        sum(size(Vt, 2) for Vt in VTs),
+    )
+
+    return Amap
+end
+
+"""
+    compute_structured_energy(projections)
+
+Given the matrix of normalized projections from `compute_normalized_projections`,
+compute the total structured energy in the whitened space:
+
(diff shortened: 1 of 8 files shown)


src/svd_analysis.jl
```diff
@@ -0,0 +1,212 @@
...
+    Avec = isa(A, AbstractVector) ? A : [A]
+
+    Us = []
+    Ss = []
+    VTs = []
+    ds = []
+    batches = []
+    shift = 0
+    for a in Avec
+        if isa(a, UniformScaling)
```
odunbar: Maybe we could add a "Diagonal" option, as this is probably the one other case that might appear somewhat regularly. Here we would just want something like
```
push!(Us, zeros)
push!(Ss, zeros)
push!(VTs, zeros)
push!(ds, a.diag)
bsize = length(a.diag)
```

src/svd_analysis.jl
```diff
@@ -0,0 +1,212 @@
...
+
+    # Reverse so index 1 = dominant eigenvector
+    idxs = n_eigenvectors:-1:1
+    eigvalues = eigvalues[idxs]
+    eigvectors = eigvectors[:, idxs]
+
+    g = EKP.get_g(ekp, iter)
+    succ_ens, _ = EKP.split_indices_by_success(g)
+    mean_g =
+        dropdims(sum(g[:, succ_ens], dims = 2), dims = 2) / length(succ_ens)
```
odunbar: Any reason why we don't use:
```julia
mean_g = mean(g[:, succ_ens], dims = 2)
```
here?

src/svd_analysis.jl
```diff
@@ -0,0 +1,212 @@
...
+where `aᵢ = vᵢᵀ diff` and the sum is over all columns of `eigvectors`.
+
+- `struct_energy > 1`: residual is dominated by structured (eigenvector) noise.
+- `struct_energy < 1`: residual is dominated by diagonal noise or high-frequency directions.
+"""
+function compute_structured_energy(diff, eigvectors)
+    a = eigvectors' * diff       # a[i] = vᵢᵀ diff
+    structured = eigvectors * a  # ∑ᵢ aᵢ vᵢ
+    unstructured = diff - structured
+    return norm(structured) / norm(unstructured)
```
odunbar: So it turns out there is a slightly better way of doing this structured energy. I think we have computed what we need for it though: namely the projections below
```
projections[i, v] = a_i_v / sqrt(eigvalues[i])...
```
The main issue is when we convert back we are currently comparing the structured energy in the original space. It actually makes more sense to keep it in the whitened space. This will then keep the interpretation with >1 and <1. 
Currently I don't think it quite does this.

The following should be correct
```
z = vec(sum(projections, dims = 2)) # sum over v
energy_in_white = sum(z[i].^2) 
normalized_energy_in_white = energy_in_white / length(z)
```
The per variable energy I think is the same as this, we just do it per v, not over the sum.

odunbar (commented):
Hi Nat,

This looks great! I think theres a better way of computing the energy which is simple and i've detailed it in the comment

src/svd_analysis.jl
odunbar: We don't need to compute the residual for the ratio either then

src/svd_analysis.jl
```diff
@@ -0,0 +1,213 @@
...
+    obs_series = EKP.get_observation_series(ekp)
+    obs_noise_cov = EKP.get_obs_noise_cov(obs_series; build = false)
+    linear_map = _create_compact_linear_map(obs_noise_cov)
+    result, _ = partialschur(linear_map; nev = n_eigenvectors)
+    eigvalues, eigvectors = partialeigen(result)
+
+    # Reverse so index 1 = dominant eigenvector
+    idxs = n_eigenvectors:-1:1
+    eigvalues = eigvalues[idxs]
+    eigvectors = eigvectors[:, idxs]
```
ph-kev:
```suggestion
    reverse!(eigvalues)
    reverse!(eigvectors, dims = 2)
```

src/svd_analysis.jl
nefrathenrici: It would technically add Statistics as a dependency but it's a stdlib so it shouldn't matter.

src/svd_analysis.jl
```diff
@@ -0,0 +1,213 @@
...
+    # TODO: Don't use g_ens_builder just to extract the metadata ranges
+    g_ens_builder = EnsembleBuilder.GEnsembleBuilder(ekp)
+    metadata_vec = g_ens_builder.metadata_vec
+    projections = compute_normalized_projections(
+        diff,
+        eigvectors,
+        eigvalues,
+        metadata_vec,
+    )
+    return (;
```
ph-kev: I am not sure if this is helpful or not, but to me, the metadata should also be returned from this function so each number can be associated to one of the metadata.

src/svd_analysis.jl
```diff
@@ -0,0 +1,213 @@
...
+
+export compute_structured_energy,
+    compute_structured_energy_by_variable,
+    compute_normalized_projections,
+    analyze_residual
+
+# TODO: Use the function in EKP once it's moved there
+"""
+    _create_compact_linear_map(A)
+
```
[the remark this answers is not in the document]
odunbar: Nat: this was Modified from CES.jl/src/Utilities.jl

src/svd_analysis.jl
```diff
@@ -0,0 +1,213 @@
...
+Given the matrix of normalized projections from `compute_normalized_projections`,
+compute the total structured energy in the whitened space:
+
+    energy = (1/n_eig) * ∑ᵢ zᵢ²,   where zᵢ = ∑ᵥ projections[i, v] = aᵢ / √λᵢ
+
+`zᵢ` is the global whitened projection onto eigenvector `i`. Under the noise
+model, `zᵢ ~ N(0, 1)`, so `energy ≈ 1` is consistent with noise. Values >> 1
+indicate mismatch beyond what the structured noise explains.
+"""
```
ph-kev, line 107: Can you include a sentence for values < 1?

src/svd_analysis.jl
```diff
@@ -0,0 +1,213 @@
...
+function compute_structured_energy(projections)
+    z = vec(sum(projections, dims = 2))
```
ph-kev: Do you need a `vec` here if you are going to apply a sum to it afterward?

src/svd_analysis.jl
```diff
@@ -0,0 +1,213 @@
...
+    compute_structured_energy_by_variable(projections)
+
+Given the matrix of normalized projections from `compute_normalized_projections`,
+compute the per-variable structured energy in the whitened space:
+
+    energy_v = (1/n_eig) * ∑ᵢ projections[i, v]²
+
+Returns a vector of length `n_variables`. Values >> 1 for variable `v` indicate
+that variable's contribution to the eigenvector projections exceeds noise-model
+predictions.
```
ph-kev: Can you include the cases when the values is <= 1?

src/svd_analysis.jl
```diff
@@ -0,0 +1,213 @@
...
+        end
+    end
+    return projections
+end
+
+"""
+    analyze_residual(ekp, iter; n_eigenvectors = 3)
+
+Analyze the model-data residual (y - G(u)) at iteration `iter` of an EKP calibration
+using the top eigenvectors of the noise covariance.
```
ph-kev, line 168: Is this only compatible with `SVDplusD` or is it compatible with all matrices? I think it would be good to include a note about that.

ph-kev (approved):

PR #291, opened 2026-03-18
Refactor module loading script for each backend

### LINKED ISSUE #287: Add `module_load_str` to `hpc_kwargs`
It seems strange to me that the `module_load_str` is tied to versions of `ClimaCalibrate` with no easy way of changing it as far as I know.

When I was trying to update the land calibration pipeline, there was a mismatch for the `climacommon` which lead to all the ensemble member runs failing because there were two different versions of Julia being used (1.10 and 1.12). The ways I found to fix this was to use the version of `climacommon` that was expected or commit type piracy and redefine `module_load_str` to use the version of `climacommon` that I want.

It would be nice to expose `module_load_str` in `hpc_kwargs` so users can change it themselves.

ph-kev:
closes https://github.com/CliMA/ClimaCalibrate.jl/issues/287

This PR refactors the functionality of creating the script of which modules to load depending on which backend to `ModuleSelector`. With this refactor, the option of choosing which `climacommon` to use is allowed and a warning is thrown if there is a possible mismatch between different `climacommon` versions.

TODO

- [ ] Potentially simplify `ModuleSelector` and the functionality (e.g. `AbstractClimaCommonOption`)
- [ ] Determine the user interface of `ModuleSelector` and how it will be used when constructing the backend

### DIFF
--- a/src/backends.jl
+++ b/src/backends.jl
@@ -527,3 +527,5 @@ backend_worker_kwargs(::Type{DerechoBackend}) = (; q = "main", A = "UCIT0011")
 backend_worker_kwargs(::Type{GCPBackend}) = (; partition = "a3")
 
 backend_worker_kwargs(::Type{<:AbstractBackend}) = (;)
+
+include("module.jl")
--- a/src/module.jl
+++ b/src/module.jl
@@ -1,0 +1,199 @@
+"""
+    AbstractClimaCommonOption
+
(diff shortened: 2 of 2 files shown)


ph-kev:
This functionality is now supported by https://github.com/CliMA/ClimaCalibrate.jl/pull/303

PR #298, opened 2026-04-13
Minor cleanup and fixes

### LINKED ISSUE #294: Start iteration from 1 and not 0
When ClimaCalibrate releases a breaking change, it would be good to start an iteration with `iteration_001` and not `iteration_000`.

ph-kev:
closes https://github.com/CliMA/ClimaCalibrate.jl/issues/294

This PR adds job status to be an enum instead of symbols, updates iteration indexing to one-based indexing, reorganize tests, and removes the Manifest.

The zero-based indexing was inconsistent with indexing used by Julia and EKP. Tests that doesn't involve using the backend are moved to GHA instead. Finally, the manifest was removed because leeping the manifest introduces unnecessary repository noise and busywork when updating dependencies.


nefrathenrici (approved):
Thank you, it's nice to finally change the initial iteration index from 0 to 1!

PR #301, opened 2026-04-16
Fix bugs in analyze_residual

### LINKED ISSUE #300: Add a test for analyze_residual
There is no test that run `analyze_residual`. There might potentially be a bug with functions being called from the `ObservationRecipe` module as functions implemented in the extension and not in src is not easily available to use.

ph-kev:
closes https://github.com/CliMA/ClimaCalibrate.jl/issues/300

The first bug was that `get_metadata_for_nth_iteration` was being qualified with `ObservationRecipe` instead of `EKPUtils`. This was not caught because the function was never exercised in the test suite. The second bug was an indexing issue when retrieving the covariance matrix with `get_obs_noise_cov`.

### DIFF
--- a/test/svd_analysis.jl
+++ b/test/svd_analysis.jl
@@ -1,8 +1,24 @@
 using Test
+import Dates
 using LinearAlgebra
-import EnsembleKalmanProcesses as EKP
-import ClimaCalibrate as CAL
 using LinearMaps
+import EnsembleKalmanProcesses as EKP
+import EnsembleKalmanProcesses.ParameterDistributions
+import ClimaCalibrate as CAL
+import ClimaCalibrate.ObservationRecipe
+import ClimaCalibrate.EnsembleBuilder
+import ClimaAnalysis
+import ClimaAnalysis.Template:
+    TemplateVar,
+    make_template_var,
+    add_attribs,
+    add_dim,
+    add_time_dim,
+    add_lon_dim,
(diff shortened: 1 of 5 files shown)


test/svd_analysis.jl
```diff
@@ -133,3 +149,92 @@ end
...
+            (no_lon_var,),
+            "2007-12-1",
+            "2008-9-1",
+        ),
+    ]
+    obs_series = EKP.ObservationSeries(
+        Dict(
+            "observations" => obs_vec,
+            "minibatcher" =>
+                CAL.minibatcher_over_samples(length(obs_vec), 1),
```
ph-kev, line 209: This doesn't work with `CAL.minibatcher_over_samples(length(obs_vec), 2)` when updating the ensemble. I am not sure if this is because I did not properly constructed the `EKP` object or if this is a bug with `EKP`.

nefrathenrici (approved):

test/svd_analysis.jl
nefrathenrici: Did you look into this more to determine if it's an EKP issue? I think we should be able to run with differently sized data in different minibatches.

ph-kev: This is an issue with EKP. See https://github.com/CliMA/EnsembleKalmanProcesses.jl/issues/557.

PR #302, opened 2026-04-17
Clean up utils file for surface flux perfect model experiment

ph-kev:
This PR cleans up the utils file for the surface flux perfect model experiment by

1. Moving all the imports and `@everywhere` blocks to the top of the file and removing unused imports.
2. Adding docstrings to the testing and plotting functions.
3. Simplifying plotting code such as replacing spread calculation with `Statistics.var`, using CairoMakie instead of Makie, and adding logging for output paths.
4. Relying less on global state by passing arguments to functions

This PR also removed a test that was not removed in an earlier PR https://github.com/CliMA/ClimaCalibrate.jl/pull/298 and fix some typos in the `EKPUtils` module.


nefrathenrici (approved):
Thanks! Do you think it makes sense to move the surface fluxes test case into the test folder? It isn't really an "experiment" and I don't think we plan to add more experiments.

PR #303, opened 2026-04-17
Add job scheduler configs for HPCBackends

### LINKED ISSUE #283: Add `HPCConfig` to unify and store `HPCBackend` struct info
The HPCBackends currently duplicate the same information. They could all wrap one type that stores the common information: `experiment_dir`, `model_interface`, `module_load_str`, `hpc_kwargs`,
`verbose`, `reruns`, `exeflags`. This will reduce code duplication and improve function signatures.

```
struct HPCConfig
    experiment_dir::String       = project_dir()
    model_interface::String      = default_model_interface()
    module_load_str::String      = ""
    hpc_kwargs::Dict{Symbol,Any} = Dict{Symbol,Any}()
    verbose::Bool                = false
    reruns::Int                  = 1
    exeflags::String             = ""
end

struct CaltechHPCBackend <: SlurmBackend
    config::HPCConfig = HPCConfig()
end
```

### LINKED ISSUE #287: Add `module_load_str` to `hpc_kwargs`

ph-kev:
closes https://github.com/CliMA/ClimaCalibrate.jl/issues/287, closes https://github.com/CliMA/ClimaCalibrate.jl/issues/283

This PR adds `SlurmConfig` and `PBSConfig` that will be part of the `HPCBackend` structs. Compared to the current implementation, the user can also specify the environment variables and modules which was a pain point for the calibration pipeline. Furthermore, there is more modularity in how the directives, modules, and environment variables are parsed, validated, and transformed to be used in the job scripts.

TODO

- [ ] ~~Add job scheduler agnostic config with mapping to `SlurmBackend` and `PBSBackend`~~
- [x] Add parsing for single character directives (e.g. `-b` instead of `--begin`)

I learned that the current job configuration for PBS assumes that you pass in directives that follows the Slurm naming conventions and when making the job script, it translates them to match the PBS naming conventions. Given this design, introducing a job scheduler agnostic config would add unnecessary complexity without benefit.


src/backends/config.jl
```diff
@@ -0,0 +1,420 @@
...
+"""
+    SlurmConfig <: AbstractHPCConfig
+
+A configuration holding Slurm directives, modules, and environment variables
+that will be used when creating a job scripts by the `SlurmBackend`s.
+"""
+struct SlurmConfig <: AbstractHPCConfig
+    directives::Vector{Pair{Symbol, Any}}
+    modules::Vector{String}
+    env_vars::Vector{Pair{String, Any}}
```
ph-kev: I wasn't too sure about this. I want to use a vector instead of a dictionary, so making the job script is deterministic. However, this does make it harder to work with. See `_add_unique_entry!` for example.

src/backend.jl
```diff
@@ -141,9 +143,13 @@ Used for Caltech's
...
-Base.@kwdef struct CaltechHPCBackend <: SlurmBackend
-    hpc_kwargs::Dict{Symbol, Any} = Dict{Symbol, Any}()
-    job_records::Vector{JobInfo} = []
+struct CaltechHPCBackend <: SlurmBackend
+    hpc_config::SlurmConfig
+    job_records::Vector{JobInfo}
+end
+
+function CaltechHPCBackend(config::SlurmConfig)
+    return CaltechHPCBackend(config, [])
```
ph-kev, line 152: Another option for the constructor is to pass directives, modules, and environment variables to `CaltechHPCBackend` instead of pass a `SlurmConfig`. This hides the details of `SlurmConfig` and `PBSConfig`, but I am not sure if it is worth hiding the detail since the behavior of both are different.

ph-kev (commented):
This PR is ready for initial review, but I want to see if it can be improved or changed before merging it in.

src/backends/config.jl
ph-kev: OrderedCollections should be used here

src/backend.jl
ph-kev: I added the constructor for the different `HPCBackend`s that makes the `PBSConfig` or `SlurmConfig` internally. The `PBSConfig` and `SlurmConfig` are still public facing though since it is easier to write the documentation in one place rather than repeating it for both the `HPCBackend`s and `AbstractHPCConfig`s.

src/backends/config.jl
```diff
@@ -0,0 +1,431 @@
...
+
+This function does not validate whether `str` is correct or not.
+"""
+format_slurm_time(str::AbstractString) = str
+
+
+"""
+    format_pbs_time(minutes::Int)
+
+Format `minutes` into a string (HH:MM:SS) accecpted by PBS.
```
nefrathenrici:
```suggestion
Format `minutes` into a string (HH:MM:SS) accepted by PBS.
```

src/backends/config.jl
```diff
@@ -0,0 +1,431 @@
...
+- `CLIMACOMMS_CONTEXT`: "MPI".
+
+## Examples
+
+This example creates a Slurm configuration for a job with a single task, using
+12 CPUs and 1 GPU, and a runtime of 720 minutes. It loads the latest version of
+climacommon and explicitly sets environment variables for ClimaComms.
+
+```julia
+ClimaCalibrate.PBSConfig(;
```
nefrathenrici:
```suggestion
ClimaCalibrate.SlurmConfig(;
```

src/backends/config.jl
```diff
@@ -0,0 +1,431 @@
...
+"""
+function generate_directives(config::PBSConfig)
+    (; directives) = config
+    queue = directives[:queue]
+    walltime = directives[:time]
+    num_nodes = directives[:ntasks]
+    cpus_per_node = directives[:cpus_per_task]
+    gpus_per_node = directives[:gpus_per_task]
+    job_priority = directives[:job_priority]
+    ranks_per_node = gpus_per_node > 0 ? gpus_per_node : cpus_per_node
```
nefrathenrici, line 350: I think it would be helpful to add a check and warn if there are unused directives.

src/backends/config.jl
```diff
@@ -0,0 +1,431 @@
...
+function _parse_config_args(directives, modules, env_vars)
+    owned_directives = OrderedDict{Symbol, Any}()
+    owned_modules = String[]
+    owned_env_vars = OrderedDict{String, Any}()
+
+    for (key, val) in directives
+        key = Symbol(key)
+        haskey(owned_directives, key) &&
+            error("Not all directives ($key) are unique")
+        owned_directives[Symbol(key)] = val
```
nefrathenrici:
```suggestion
        owned_directives[key] = val
```

nefrathenrici (approved):
Thank you! This is structured a lot better. Could you update this [documentation page](https://github.com/CliMA/ClimaCalibrate.jl/blob/main/docs/src/submit_scripts.md) or open an issue for it?

PR #312, opened 2026-04-30
Add model interface abstract type

### LINKED ISSUE #281: Improve user-overriden functions
Three current functions are provided by the user as hooks in the calibration lifecycle: `forward_model`, `observation_map`, `analyze_iteration`.

The information passed into these functions is quite limited, and pipelines commonly end up loading the EKP struct and accessing global variables to compensate. We should provide a data structure containing these to each function to ensure that they have the needed variables by default.

Rough sketch of the data structure:
```
struct CalibrationContext
    iteration::Int
    member::Int
    output_dir::String
    ensemble_size::Int
    prior::Union{ParameterDistribution, Nothing}
    ekp::Union{EKP.EnsembleKalmanProcess, Nothing}
end
```

New function calls will be:
- `forward_model(ctx::CalibrationContext)`
- `observation_map(ctx::CalibrationContext)`
- `analyze_iteration(ctx::CalibrationContext, g_ensemble)`

ph-kev:
closes https://github.com/CliMA/ClimaCalibrate.jl/issues/281

This PR adds the abstract type `AbstractModelInterface` that the user can subtype. This removes the type piracy that the user needed to do before and allows for the user to implement multiple concrete types of an `AbstractModelInterface` for supporting multiple calibration. For instance, in the coupler calibration pipeline, both subseasonal and AMIP calibration can be support with the type hierarchy
1. `AMIPModelInterface <: AbstractCouplerModelInterface`,
2. `SubseasonalModelInterface <: AbstractCouplerModelInterface`,
3. `AbstractCouplerModelInterface <: AbstractModelInterface`.

The `AbstractCouplerModelInterface` can provide default for the forward model, but the observation map can differ between the `AMIPModelInterface` and `SubseasonalModelInterface`.

### DIFF
--- a/docs/literate_example.jl
+++ b/docs/literate_example.jl
@@ -1,9 +1,9 @@
 # # Distributed Calibration Tutorial Using Julia Workers
 # This example will teach you how to use ClimaCalibrate to parallelize your calibration with workers.
-# Workers are additional processes spun up to run code in a distributed fashion. 
+# Workers are additional processes spun up to run code in a distributed fashion.
 # In this tutorial, we will run ensemble members' forward models on different workers.
 
-# The example calibration uses CliMA's atmosphere model, [`ClimaAtmos.jl`](https://github.com/CliMA/ClimaAtmos.jl/), 
+# The example calibration uses CliMA's atmosphere model, [`ClimaAtmos.jl`](https://github.com/CliMA/ClimaAtmos.jl/),
 # in a column spatial configuration for 30 days to simulate outgoing radiative fluxes.
 # Radiative fluxes are used in the observation map to calibrate the astronomical unit.
 
@@ -15,7 +15,7 @@ using ClimaUtilities.ClimaArtifacts
 import EnsembleKalmanProcesses as EKP
 import EnsembleKalmanProcesses: I, ParameterDistributions.constrained_gaussian
 
-# Next, we add workers. These are primarily added by 
+# Next, we add workers. These are primarily added by
 # [`Distributed.addprocs`](https://docs.julialang.org/en/v1/stdlib/Distributed/#Distributed.addprocs)
 # or by starting Julia with multiple processes: `julia -p <nprocs>`.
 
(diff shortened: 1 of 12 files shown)


docs/literate_example.jl
```diff
@@ -63,8 +63,24 @@ mkpath(output_dir)
...
+#     Due to limitations in Documenter.jl (see
+#     [here](https://github.com/fredrikekre/Literate.jl/issues/254) and
+#     [here](https://github.com/JuliaDocs/Documenter.jl/issues/1848)), we append
+#     `@eval $(@__MODULE__)` to every `@everywhere` call. For your own
+#     calibration script, you do not need to do this.
+
+@everywhere @eval $(@__MODULE__) struct RadiativeFluxModelInterface <:
+                                        CAL.AbstractModelInterface end
+
+@everywhere @eval $(@__MODULE__) function CAL.forward_model(
```
ph-kev, line 79: For this, the issues link toward some possible solutions. I want to try them first to see if it improves the doc.

src/calibration.jl
```diff
@@ -17,10 +17,11 @@ include("ekp_interface.jl")
...
         backend::HPCBackend,
         ekp::EKP.EnsembleKalmanProcess,
+        interface::AbstractModelInterface,
         n_iterations,
         prior,
         output_dir,
-        model_interface;
+        model_interface_filepath;
         experiment_dir = project_dir(),
         exeflags = "",
```
ph-kev: Would it make sense to add  `model_interface_filepath`, `experiment_dir`, and `exeflags` passed by the model interface instead? For example, the user would define `ClimaCalibrate.model_interface_filepath(::MyModelInterface) = ...`, `ClimaCalibrate.experiment_dir(::MyModelInterface) = ...` , and `ClimaCalibrate.exeflags(::MyModelInterface) = ...`?

I am worried that this would be overloading the `ModelInterface` struct, but this would keep the interface between all of the calibrate functions the same.

src/model_interface.jl
```diff
@@ -1,32 +1,78 @@
 import EnsembleKalmanProcesses as EKP
```
ph-kev, line 1: This could be its own module. I am not sure what is a good name for these collection of functions.

ph-kev:
There are still some issues that need to be resolved, but they are minor. The updated land calibration pipeline can be found [here](https://github.com/CliMA/ClimaLand.jl/tree/kp/update-calibration) and the updated coupler calibration pipeline can be found [here](https://github.com/CliMA/ClimaCoupler.jl/tree/kp/update-calibration).

src/calibration.jl
nefrathenrici: I think it makes sense to turn these into methods instead of kwargs, but they should be used uniformly by all backends.

nefrathenrici: I think at minimum model_interface_filepath should be a method and not a kwarg.

src/model_interface.jl
nefrathenrici: I don't think we need this to be its own module yet, if we add more functionality or helper methods we should reconsider.

src/calibration.jl
ph-kev: I made all the keyword arguments into functions. The only thing I dislike about this is that `experiment_dir` and `exeflags` are defined with the model interface, but the `WorkerBackend` doesn't use that and rely on `addprocs` to pass `exeflags` to the workers and the default for the project directory can't be changed.

docs/literate_example.jl
```diff
@@ -86,19 +102,19 @@ mkpath(output_dir)
             ),
         ],
     )
-    #md # Set the output path for the current member
+    ## # Set the output path for the current member
```
nefrathenrici: Not sure why this change is needed, it should be fine either way.

src/calibration.jl
nefrathenrici: On second thought, it might be cleaner to move the `experiment_dir` and `exeflags` into the `SlurmConfig`/`PBSConfig`, since they're really backend concerns and not model-specific. Do you think that makes sense?

src/calibration.jl
```diff
@@ -17,10 +17,11 @@ include("ekp_interface.jl")
     calibrate(
```
nefrathenrici, line 17: This docstring needs to be updated after we decide what to do with the `experiment_dir` and `exeflags` options.

nefrathenrici (approved):
Thank you, this looks good! The `exeflags` and `experiment_dir` options still need to be sorted out.

src/calibration.jl
ph-kev: I think that might be a little better. Currently, the configs doesn't care about what the script is running. These settings are only specific to running Julia scripts which is why it doesn't feel right to me.

nefrathenrici: You're right, the HPCConfig should not deal with Julia-specific details. Let's keep it separate. The documentation makes it clear that these methods don't always apply.

PR #317, opened 2026-05-26
Set GPU rank for PBS jobs

nefrathenrici:
This PR sets GPU ranks for PBS jobs via a bash script adapted from Derecho's `/opt/utils/bin/set_gpu_rank`.

We currently use the external `/opt/utils/bin/set_gpu_rank` script on Derecho to set GPU ranks.

The script hosted on Derecho runs `exec $*` at the end which word-splits and drops the quoting on the `-e` body . The call `set_gpu_rank julia --project=... -e import ClimaCalibrate; ...` gets parsed as `julia --project.. -e import ClimaCalibrate;...` so julia just evaluates `import` and errors. See [build](https://buildkite.com/clima/climacoupler-calibration/builds/11#019e5718-18ef-479c-b68d-9a3a7f6a4a4a) and error below. 

The script added here runs `exec "$@"` instead, which does not drop the quotes.
```
│ set_gpu_rank: using cray-mpich MPI library with 1 accelerators per node on 1 hosts...
│ set_gpu_rank: setting MPICH_GPU_SUPPORT_ENABLED=1, MPICH_OFI_NIC_POLICY=GPU
│ set_gpu_rank: host deg0038 / global rank 0 / local rank 0 / CUDA_VISIBLE_DEVICES 0
│ ERROR: ParseError:
│ # Error @ none:1:7
│ import
│ #     └ ── premature end of input
│ Stacktrace:
│  [1] top-level scope
│    @ none:1
│  [2] eval(m::Module, e::Any)
│    @ Core ./boot.jl:489
│  [3] exec_options(opts::Base.JLOptions)
│    @ Base ./client.jl:283
│  [4] _start()
│    @ Base ./client.jl:550
│ deg0038.hsn.de.hpc.ucar.edu: rank 0 exited with code 1
└ @ ClimaCalibrate.Calibration /glade/derecho/scratch/nefrathe/slurm-buildkite/climacoupler-calibration/11/depot/default/packages/ClimaCalibrate/7Y2sG/src/calibration.jl:317
```

### DIFF
--- a/src/backends/pbs.jl
+++ b/src/backends/pbs.jl
@@ -26,7 +26,9 @@ function make_job_script(
 
     if gpus_per_node > 0
         ranks_per_node = gpus_per_node
-        set_gpu_rank = "set_gpu_rank"
+        # Use a bash script to set GPU ranks for each process, needed so that 
+        # MPI can properly use multiple GPUs concurrently
+        set_gpu_rank = joinpath(@__DIR__, "set_gpu_rank.sh")
     else
         ranks_per_node = cpus_per_node
         set_gpu_rank = ""
@@ -191,7 +193,8 @@ end
 
 function module_load_string(backend::DerechoBackend)
     module_loads = generate_modules(backend.hpc_config)
-    return """export MODULEPATH="/glade/campaign/univ/ucit0011/ClimaModules-Derecho:\$MODULEPATH"
-    module purge
-    $module_loads"""
+    return """module purge
+    module use /glade/campaign/univ/ucit0011/ClimaModules-Derecho
+    $module_loads
+    module list 2>&1"""
 end
--- a/src/backends/set_gpu_rank.sh
+++ b/src/backends/set_gpu_rank.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+#
+# Maps MPI ranks to GPUs for ClimaCalibrate's PBS backend.
+#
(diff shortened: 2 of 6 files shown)


src/backends/set_gpu_rank.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+#
+# Maps MPI ranks to GPUs for ClimaCalibrate's PBS backend.
+#
+# This is adapted from NCAR Derecho's /opt/utils/bin/set_gpu_rank.
```
ph-kev, line 5: Is this something that we need to keep updated if the .sh file on Derecho change?

src/backends/pbs.jl
```diff
@@ -26,7 +26,7 @@ function make_job_script(
 
     if gpus_per_node > 0
         ranks_per_node = gpus_per_node
-        set_gpu_rank = "set_gpu_rank"
+        set_gpu_rank = joinpath(@__DIR__, "set_gpu_rank.sh")
```
ph-kev, line 31: Can you a note on why we need to do this or add a note that link to this PR?

ph-kev (approved):

src/backends/set_gpu_rank.sh
nefrathenrici: No, we should not need to change this except to update the supported MPI libraries.

PR #320, opened 2026-06-02
Add support for `OutputVar`s with no time dimension

### LINKED ISSUE #267: Add support for OutputVars with no time dimension in extension
For calibration using ClimaAnalysis as a data processing backend, there is an implicit assumption that all the `OutputVar`s contain a time dimension. It would be nice to remove this assumption and be able to run a calibration for `OutputVar`s with no time dimension.

You might get a `OutputVar` with no time dimension when you do an average of the time dimension which remove the time dimension.

ph-kev:
For `ObservationRecipe` and `EnsembleBuilder`, there was an implicit assumption that calibration would always involve time series data. However, this may not always be the case. For example, instead of using the entire time series data, you may be interested in certain features of the time series data (e.g. where the minimum or maximum are or the amplitude if it is periodic data). This PR allows for `OutputVar`s with no time dimension to be used with the `GEnsembleBuilder` and `ObservationRecipe` with `ScalarCovariance`.

In the future, we may want to support `OutputVar`s with no time dimension for the `SVDplusDCovariance` estimator, but this can be done in another PR.

closes https://github.com/CliMA/ClimaCalibrate.jl/issues/267

# TODO

- [x] Reduce code duplication for `covariance` and `observation` in `ObservationRecipe` for `ScalarCovariance`

### DIFF
--- a/ext/observation_recipe.jl
+++ b/ext/observation_recipe.jl
@@ -443,7 +443,7 @@ function ObservationRecipe.observation(
     # Get the flattened sample and metadata
     windowed_vars =
         ClimaAnalysis.window.(vars, "time", left = start_date, right = end_date)
-    flat_vars = map(var -> ClimaAnalysis.flatten(var), windowed_vars)
+    flat_vars = map(var -> ClimaAnalysis.flatten(var; dims), windowed_vars)
     stacked_sample = vcat((flat_var.data for flat_var in flat_vars)...)
     metadata = [(flat_var.metadata for flat_var in flat_vars)...]
     covar = ObservationRecipe.covariance(
--- a/ext/observation_recipe.jl
+++ b/ext/observation_recipe.jl
@@ -28,23 +28,66 @@ function ObservationRecipe.covariance(
     end_date;
     dims = FLATTENED_DIMS,
 )
-    # Convert dates to Dates.DateTime if they are strings
-    start_date = Dates.DateTime(start_date)
-    end_date = Dates.DateTime(end_date)
+    return _covariance(covar_estimator, vars, start_date, end_date; dims)
+end
 
+"""
+    covariance(
+        covar_estimator::ScalarCovariance,
+        vars::Union{OutputVar, Iterable{OutputVar}};
+        dims = $FLATTENED_DIMS,
+    )
+
+Compute the scalar covariance matrix.
+
+Data from `vars` will not be used to compute the covariance matrix.
+"""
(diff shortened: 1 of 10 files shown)


ext/observation_recipe.jl
```diff
@@ -65,6 +65,44 @@ function ObservationRecipe.covariance(
...
+                ClimaAnalysis.flatten(
+                    _lat_weights_var(
+                        var,
+                        min_cosd_lat = covar_estimator.min_cosd_lat,
+                    );
+                    dims,
+                ).data
+            diag_cov .*= flattened_lat_weights
+        end
+        diag_cov
```
nefrathenrici, line 106:
```suggestion
          _scalar_diag(covar_estimator, var; dims)
```
I think you can move this duplicated code into a helper function.
```
function _scalar_diag(covar_estimator::ScalarCovariance, var; dims)
    flattened_data = ClimaAnalysis.flatten(var; dims).data
    diag_cov = ones(eltype(flattened_data), size(flattened_data)...)
    diag_cov .*= covar_estimator.scalar

    if covar_estimator.use_latitude_weights
        flattened_lat_weights =
            ClimaAnalysis.flatten(
                _lat_weights_var(
                    var,
                    min_cosd_lat = covar_estimator.min_cosd_lat,
                );
                dims,
            ).data
        diag_cov .*= flattened_lat_weights
    end
    return diag_cov
end
```

ext/observation_recipe.jl
nefrathenrici:
```suggestion
        _scalar_diag(covar_estimator, var; dims)
```

if you like the helper function in the other comment it should be applied here as well.

ext/observation_recipe.jl
```diff
@@ -471,6 +509,57 @@ function ObservationRecipe.observation(
...
+        )
+    )
+    return EKP.Observation(
+        Dict(
+            "samples" => stacked_sample,
+            "covariances" => covar,
+            "names" => name,
+            "metadata" => metadata,
+        ),
+    )
```
nefrathenrici, line 558: This snippet could also be moved into a helper function.

nefrathenrici (approved):
Looks correct, I just had a comment on the code duplication.

nefrathenrici (approved):
This looks great. To make the Derecho pipeline work, could you update line 5 of `.buildkite/derecho_pipeline.yml`? It needs a GPU:
```
  pbs_l_select: "1:ncpus=1:ngpus=1"
```

PR #325, opened 2026-06-04
Remove packages from targets and extras

ph-kev:
This commit removes OrderedCollections and Statistics from the targets and extras list in the compat because they are dependencies in src now.

I noticed this from looking at the dependabot PRs.


nefrathenrici (approved):
Thanks!

PR #331, opened 2026-06-15
Add plotting utilities for calibration

ph-kev:
This PR adds plotting utilities for calibration by allowing users to plot columns of the G ensemble matrix, the mean forward map evaluation, and the observations for any given iteration as line plots.

TODO

- [x] Add docs
- [x] Add tests

For the plotting utilities added here, they do not assume that `ObservationRecipe` was used to make the observations. If this is the case, there is more potential for better plotting utilities. For example, you can denote regions of the line plots that correspond to a variable and plot the short names from the metadata in the observations. This could be done in this PR or another PR.

### DIFF
--- a/ext/ClimaCalibrateMakieExt.jl
+++ b/ext/ClimaCalibrateMakieExt.jl
@@ -1,0 +1,174 @@
+module ClimaCalibrateMakieExt
+
+import Makie
+import EnsembleKalmanProcesses as EKP
+import ClimaCalibrate
+
+import ClimaCalibrate.Visualization:
+    plot_g, plot_g!, plot_g_mean, plot_g_mean!, plot_obs, plot_obs!
+
+"""
+    plot_g
+
+Plot members of the G ensemble matrix as line plots.
+
+If the `iter` keyword argument is not passed, then this plot the last completed
+G ensemble matrix. Otherwise, it plots the `iter`th G ensemble matrix.
+"""
+Makie.@recipe Plot_G (ekp,) begin
+    "Iteration of G ensemble matrix to plot. If `nothing`, then use the last
+    iteration"
+    iter = nothing
+    Makie.documented_attributes(Makie.Lines)...
+end
+
+"""
+    Makie.plot!(g_ens_plot::Plot_G)
+
+Plot members of the G ensemble matrix as line plots.
+
+This function is called when using `Visualization.plot_g` or
+`Visualization.plot_g!`.
+"""
+function Makie.plot!(g_ens_plot::Plot_G)
+    ekp = g_ens_plot.ekp[]
+    iter = _iter_or_total(ekp, g_ens_plot.iter[])
+    g_ens_matrix = EKP.get_g(ekp, iter)
+    for j in axes(g_ens_matrix, 2)
+        ensemble_member = view(g_ens_matrix, :, j)
+        Makie.lines!(
+            g_ens_plot,
+            g_ens_plot.attributes,
+            1:length(ensemble_member),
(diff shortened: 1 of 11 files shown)


src/visualization.jl
```diff
@@ -0,0 +1,67 @@
...
+
+"""
+    plot_g!
+
+This is the mutating variant of the plotting function [`plot_g`](@ref).
+"""
+function plot_g! end
+
+"""
+    plot_g
```
nefrathenrici:
```suggestion
    plot_g_mean
```

src/visualization.jl
```diff
@@ -0,0 +1,67 @@
+module Visualization
+
+export plot_g, plot_g!, plot_g_mean, plot_g_mean!
```
nefrathenrici:
```suggestion
export plot_g, plot_g!, plot_g_mean, plot_g_mean!, plot_obs, plot_obs!
```

ext/ClimaCalibrateMakieExt.jl
```diff
@@ -0,0 +1,178 @@
...
+
+Plot the observations as a line plot.
+
+This function is called when using `Visualization.plot_obs` or
+`Visualization.plot_obs!`.
+"""
+function Makie.plot!(obs_plot::Plot_Obs)
+    ekp = obs_plot.ekp[]
+    iter = _iter_or_total(ekp, obs_plot.iter[])
+    obs_series = EKP.get_observation_series(ekp)
```
nefrathenrici, line 114: I am not familiar with Makie recipes, but it seems like we have the `convert_arguments` hook below that extracts the observation series from the EKP, but we extract the obs series again in this function. Are the calls to `get_observation_series` duplicated? Is this code path tested?

test/visualization.jl
```diff
@@ -0,0 +1,126 @@
...
+    )
+
+    iters = 3
+    for i in 1:iters
+        g_ens_builder = EnsembleBuilder.GEnsembleBuilder(ekp)
+        for i in 1:EKP.get_N_ens(ekp)
+            EnsembleBuilder.fill_g_ens_col!(g_ens_builder, i, Float64(i))
+        end
+        g_ens = EnsembleBuilder.get_g_ensemble(g_ens_builder)
+        g_ens .+= i
```
nefrathenrici, line 75:
```suggestion
        g_ens .+= j
```

test/visualization.jl
```diff
@@ -0,0 +1,126 @@
...
+        obs_series,
+        EKP.TransformUnscented(prior, impose_prior = true),
+        verbose = true,
+        scheduler = EKP.DataMisfitController(on_terminate = "continue"),
+    )
+
+    iters = 3
+    for i in 1:iters
+        g_ens_builder = EnsembleBuilder.GEnsembleBuilder(ekp)
+        for i in 1:EKP.get_N_ens(ekp)
```
nefrathenrici:
```suggestion
        for j in 1:EKP.get_N_ens(ekp)
```

nefrathenrici: Duplicate variable name in the nested loop

nefrathenrici (approved):
Thanks! I had one major concern about the use of `convert_arguments`, the rest are nits.

ext/ClimaCalibrateMakieExt.jl
ph-kev: I am not sure why, but the `convert_arguements` function was never called. It was also something that should be removed, since to get the final iteration number, we need the `EnsembleKalmanProcess` object which have that information.

PR #333, opened 2026-06-24
Fix documentation issues and minor bugs

ph-kev:
I asked Claude to go through ClimaCalibrate and ask it to find issues and bugs with the documentation and functions in `src` and `ext` and tests.

### DIFF
--- a/src/backends/pbs.jl
+++ b/src/backends/pbs.jl
@@ -90,10 +90,7 @@ function submit_job(backend::DerechoBackend, job_script::String)
             cmd = `qsub -V $pbs_filepath`
             # readchomp return a substring
             job_id = String(readchomp(setenv(cmd, clean_env)))
-            # Parse job ID, handling potential format issues
-            if isnothing(job_id)
-                error("Failed to parse job ID from output: $job_id")
-            end
+            isempty(job_id) && error("empty job ID returned")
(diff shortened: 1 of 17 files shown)


src/backends/pbs.jl
```diff
@@ -90,10 +90,7 @@ function submit_job(backend::DerechoBackend, job_script::String)
             cmd = `qsub -V $pbs_filepath`
             # readchomp return a substring
             job_id = String(readchomp(setenv(cmd, clean_env)))
-            # Parse job ID, handling potential format issues
-            if isnothing(job_id)
-                error("Failed to parse job ID from output: $job_id")
-            end
+            isempty(job_id) && error("Failed to submit PBS job: empty job ID returned")
```
nefrathenrici: This is a nit since the code works fine, but while we're fixing this, the error should probably be surfaced after the try/catch, since it will be caught below

nefrathenrici (approved):
Thank you!

src/backends/pbs.jl
ph-kev: I decided to update it to

```julia
isempty(job_id) && error("empty job ID returned")
```

which should give the right error message instead of `Failed to submit PBS job: Failed to submit PBS job: empty job ID returned".

PR #334, opened 2026-06-30
Refactor sampling component from ObservationRecipe into SampleBuilder

ph-kev:
closes #328 - This PR refactors the sampling part from `ObservationRecipe` into `SampleBuilder`. 

All relevant functions in `ObservationRecipe` takes in a `ObservedSampleCollection` which does all the checks of the `OutputVar`s. This eliminates most of the checks from `ObservationRecipe` and move them to `SampleBuilder`.

The `SampleBuilder` module features a `SampleCollection` struct and an `ObservedSampleCollection` struct. A `SampleCollection` struct features a matrix of samples and its associated metadata and an `ObservedSampleCollection` struct also includes which sample will be the observation in the calibration. These two structs are separated to allow for future changes to how the observation is chosen without breaking changes.

For the future, we might be interested in including `NaN`s in the matrix of samples and letting `ObservationRecipe.observation` handle the `NaN`s. In addition, we might also want `hcat` and `vcat` to be define for `ObservedSampleCollection` to allow for more flexibility in constructing this object.

TODO

- [x] Add to NEWS.md
- [x] Update ObservationRecipe docs
- [x] Check tests


docs/src/api.md
```diff
@@ -110,6 +110,23 @@ ClimaCalibrate.EKPUtils.get_metadata_for_nth_iteration
...
 
+## Sample Builder Interface
+
+```@docs
+ClimaCalibrateClimaAnalysisExt.SampleCollection
+ClimaCalibrate.SampleBuilder.generate_samples
+ClimaCalibrate.SampleBuilder.generate_samples_by_times
+ClimaCalibrate.SampleBuilder.num_samples
+ClimaCalibrate.SampleBuilder.reconstruct_col
+ClimaCalibrateClimaAnalysisExt.ObservedSampleCollection
```
ph-kev: I don't like the names `SampleCollection` and `ObservedSampleCollection` and I think there are probably better names for these two structs. I am against names that include the word "matrix" in them since the structs don't implement the `AbstractArray` interface.

docs/src/sample_builder.md
```diff
@@ -0,0 +1,199 @@
...
+2. the flattened vector sizes are the same,
+3. the units are the same,
+4. the dimensions between both are the same,
+5. the number of dimensions are the same,
+6. the dimension units are the same,
+7. the dimension values are the same,
+8. the coordinates where the NaNs are dropped are the same.
+
+You can exclude dimensions from these checks with the `ignore_dims` keyword
+argument. This is useful when the samples are meant to differ along a dimension.
```
ph-kev, line 96: This feature might be needed for the `GEnsembleBuilder` but it is not needed right now.

NEWS.md
```diff
@@ -13,6 +13,31 @@ main
...
+    `ClimaAnalysis.OutputVar`s into a matrix of samples with metadata.
+  - Building an observation is now a two-step process: use `SampleBuilder` to
+    turn `ClimaAnalysis.OutputVar`s into an `ObservedSampleCollection`, then
+    pass that to `ObservationRecipe` to estimate the covariance and build the
+    `EKP.Observation`.
+  - **Breaking**: `ObservationRecipe.observation` and
+    `ObservationRecipe.covariance` now take an `ObservedSampleCollection` (e.g
+    `observation(covar_estimator, osc)`) instead of
+    `(covar_estimator, vars, start_date, end_date)`. The `dims` keyword (flatten
+    order) is moved to `SampleBuilder.generate_samples`.
```
nefrathenrici: This needs to be updated

ext/sample_builder.jl
```diff
@@ -0,0 +1,553 @@
...
+    observation."""
+    i::Int
+end
+
+"""
+    choose_obs(sample_collection::SampleCollection, i::Integer)
+
+Choose the `i`th sample to be the observation for the calibration.
+"""
+function SampleBuilder.choose_obs(
```
nefrathenrici: I think this function could be folded into `ObservationRecipe.observation` with the new signature `ObservationRecipe.observation(covar_estimator, sc, i=1)`:
```julia
# before
osc = SampleBuilder.choose_obs(
    SampleBuilder.build_samples_by_times(vars, ranges; FT = Float32), 1)
obs = ObservationRecipe.observation(covar_estimator, osc)

# after
sc  = SampleBuilder.build_samples_by_times(vars, ranges; FT = Float32)
obs = ObservationRecipe.observation(covar_estimator, sc)  # obs column 1
# or pick a different column explicitly:
obs = ObservationRecipe.observation(covar_estimator, sc, 2) # obs column 2
```

nefrathenrici (approved):
Thank you, `covariance` and `observation`  are much cleaner now!

docs/src/sample_builder.md
```diff
@@ -0,0 +1,199 @@
...
+    To enable this module, use `using ClimaAnalysis` or `import
+    ClimaAnalysis`.
+
+## Why use SampleBuilder?
+
+For calibration of climate models using EnsembleKalmanProcesses.jl and other
+machine learning algorithms, observations are represented as a vector of numbers
+along with a covariance matrix describing the uncertainty of the observation.
+However, the diagnostics of climate models are typically in the form of gridded
+fields with named dimensions such as latitude, longitude, and time.
```
odunbar: I think maybe it would be more helpful to say "On the other hand, diagnostics from ClimaAnalysis are stored as "OutputVars", which store them as e.g. ...(be more explicit)

docs/src/sample_builder.md
```diff
@@ -0,0 +1,199 @@
...
+
+The `SampleBuilder` module bridges this gap. It flattens one or more
```
odunbar: "...to bridge this gap, three operations need to be performed. 
1.  we flatten each `OutputVar` into a vector, and store information for the flattening/unflattening operation. This is implemented by `SamplesBuilder.build_samples` constructing a `SampleBuilder.SampleCollection`. 
2.  we construct the associated covariance matrix for the flattened samples. This is typically prescribed or estimated from the flattened samples. This is dispatched by defining `<: ObservationRecipe.AbstractCovarianceEstimator` types
3. we combine these in an `EKP.Observation`. Implemented by the `ObservationRecipe.observation` "

docs/src/sample_builder.md
```diff
@@ -0,0 +1,199 @@
...
+`ClimaAnalysis.OutputVar`s into a matrix of samples along with their associated
+metadata and packages the result as a [`SampleCollection`](@ref
+ClimaCalibrateClimaAnalysisExt.SampleCollection). From that matrix you pick one
+column as the true observation to create an [`ObservedSampleCollection`](@ref
+ClimaCalibrateClimaAnalysisExt.ObservedSampleCollection). The
```
odunbar: The `ObservedSampleCollection` feels a bit cumbersome, i can't see why it is necessary to have this as an actual object. What you choose to take as an observation is a property of the experiment, not a property of the samples. For example,  would it not feel more natural for this to be passed in as a third argument to `ObsRecipe.observation`?

i'm also thinking about things like how would minibatching fit into your framework here?

docs/src/sample_builder.md
```diff
@@ -0,0 +1,199 @@
...
+`OutputVar`, a vector of `OutputVar`s, or a matrix of `OutputVar`s and produces
+a `SampleCollection` from it. For a single `OutputVar`, this is interpreted as a
+single sample consisting of a single `OutputVar`, and for a vector of
+`OutputVar`s, this is interpreted as a single sample consisting of multiple
+`OutputVar`s. You can think of the latter case as the vector being a column
+vector.
+
+```math
+\begin{pmatrix}
+\mathrm{OV}_{11} & \cdots & \mathrm{OV}_{1m} \\
```
odunbar: I'd use `V` for the `OutputVar` not "OV" as in mathmode this looks like a product of two variables. also the "math" term for flatten is vec. e.g. 

$$v_{ij} = \mathrm{vec}(V_{ij})$$

odunbar: I am a bit hung up on the the matrix idea. Would it be correct in thinking that really this is basically just a guard to ensure that the user provides a sample of each variable? If so, when presenting the math and describing the structure we should use maybe the following notations:

$$
\begin{pmatrix}
(\mathrm{V}^{(1)})_1 & \cdots & (\mathrm{V}^{(1)})_m\\
\vdots  & \ddots & \vdots \\
(\mathrm{V}^{(n)})_1 & \cdots &  (\mathrm{V}^{(n)})_m \end{pmatrix}
\rightarrow
\left(\begin{pmatrix}
(\mathrm{v}^{(1)})_1 \\
\vdots   \\
(\mathrm{v}^{(n)})_1 \end{pmatrix}, \dots, 
\begin{pmatrix}(\mathrm{v}^{(1)})_m  \\
\vdots \\
(\mathrm{v}^{(n)})_m\end{pmatrix}
\right)
$$

This clearly indicates  `n` variable types and `m` samples of these variables. Removing the ambiguity of the dimensions, and highlighting why we would have a "matrix" of `OutputVar`s in the first place.

docs/src/api.md
odunbar: How about `StackedSamples` or `FlatStackedOutputVars`. as a couple more options... for samples collection

docs/src/sample_builder.md
```diff
@@ -0,0 +1,199 @@
...
+SampleBuilder.build_samples_by_times(
+    [pr_var, rsut_var],
+    [(0.0, 1.0), (2.0, 3.0), (4.0, 5.0)], # this also works with Dates.DateTime
+)
+```
+
+There are also getter functions such as [`get_obs`](@ref) for getting the
+observation for the calibration, [`get_obs_metadata`](@ref) for getting the
+metadata of the observation, [`get_samples`](@ref) for getting the matrix of
+samples, [`get_metadata`](@ref) for getting the matrix of metadata.
```
odunbar: It feels like really when dealing with the `SamplesCollection` objects we want to only get information about a sample or set of samples, (e.g. get a sample, get the metadata for the stacked sample etc.). But as we store it internally as a matrix - it leads to the need to do column indexing and return of matrix values. Would it not be cleaner to store internally as a vector of (stacked-) vectors?

odunbar (changes requested):
Hi @ph-kev This is great. 

I have a few comments - that I would be keen to discus.
- Do we need a separate module for Samples builder?
- I'm not sure if we actually need a separate object for ObservaionCollection, is this actually useful? Or should it just be another argument in `observation(...)` 
- I think the docs, and possibly the internal storage too, should separate between a "vector of (stacked-) vectors" and a matrix. As it seems we end up with needing lots of matrix return methods to interrogate the samples collection objects despite them fundamentally not being a matrix any more.

docs/src/sample_builder.md
ph-kev: I feel like dealing with vector of stacked vector would be annoying since it introduces an extra thing to validate. Also, it makes the process of making the covariance matrix (for some of them) easier. I don't see the advantage of storing it as a vector of stacked-vectors.

docs/src/sample_builder.md
ph-kev: The phrasing of your sentence is confusing to me. I would say the diagnostics produced from a simulation can be processed with ClimaAnalysis which store them as `OutputVar`s. I can add a sentence or two about this.

docs/src/sample_builder.md
ph-kev: The organization of the structs is just to make it easier for me to validate everything. Also, the current use case is that one of the samples is chosen as the observational data so this just makes it easier for the users by providing the function `SampleBuilder.choose_obs(sample_collection::SampleCollection, i::Integer)`. I can make it a third argument, but I don't think it lead well to future changes.

For minibatching, the only minibatching that I seen so far is minibatching across the time dimension which in this case, you would make multiple `EKP.Observation` for each time range and minibatch like that.

docs/src/sample_builder.md
ph-kev: Yes, the matrix is just to guard against that. Before, I tried using a vector of vectors of `OutputVar`s which is painful to deal with. We can update the notation to that.

docs/src/sample_builder.md
odunbar: If it's annoying for you thats fine to leave as is;

 I just wanted to highlight that we are putting things in a matrix, but then defining all of our functions and operations which treat this matrix as a list of vectors (thus everything becomes columnwise). however if this is not an overhead, but actually a convenience then I'm happy with that

odunbar (commented):
Thanks for the response! 

I'm happy to concede to your judgement on
- Matrix vs Vec of Vecs.
- Having the additional module, 

I think my only pushback is on the `ObservedSampleCollection`.
- From what I can see `ObservedSampleCollection`'s _only_ use is to be passed into `ObsRecipe.observation(..)`. Why i think it's a bit cumbersome is that e.g. for minibatching, one needs to create an `ObsRecipe.observation(..)` for each sample in the `SampleCollection`, so now this also will require users creating an `ObservedSampleCollection` for each sample in the `SampleCollection` first to be passed in.  I think it's fine to have this layer _if_ you use the `ObservedSampleCollection` elsewhere; but otherwise it's hard to see why you would need this over a function that just passes in an index `i` to `ObsRecipe.observation(sample_collection,..., i)`. If you do actually use it for other things, it would also be nice to highlight this somewhere, because it was not clear to me at least.

Otherwise, this is a great improvement and happy to see it go in!

ext/sample_builder.jl
```diff
@@ -0,0 +1,467 @@
...
+        end
+    end
+    return SampleCollection(samples_mat, metadata_mat)
+end
+
+"""
+    change_data_type(var, FT)
+
+Return `var` with its data converted to `FT`.
+"""
```
ph-kev, line 156: I tested this in ClimaLand and I noticed the file size of the saved observation vector is bigger when this PR. The reason is because the data are saved in Float64. As a result, the metadata of the flattened `OutputVar` are also saved in `Float64` which cause a considerable increase in storage size. To fix this, we just convert the data here.

ext/sample_builder.jl
```diff
@@ -0,0 +1,467 @@
...
+)
+    @static if pkgversion(ClimaAnalysis) > v"0.5.22"
+        return ClimaAnalysis.arecompatible(
+            metadata,
+            flattened_var.metadata;
+            ignore_dims,
+        )
+    else
+        return metadata.drop_mask == flattened_var.metadata.drop_mask
+    end
```
ph-kev: I am not sure if it is worth keeping backward compatibility with older versions of ClimaAnalysis here.

odunbar (approved):
LGTM! Thanks for the update, a great improvement towards much-needd flexible observations!

ext/utils.jl
```diff
@@ -1,149 +1,78 @@
...
-    season_to_month = Dict("MAM" => 3, "JJA" => 6, "SON" => 9, "DJF" => 12)
-    first_date_of_season =
-        Dates.DateTime(first_year, season_to_month[first_season], 1)
-    curr_date = first_date_of_season
-    seasons = String[]
-    while curr_date <= end_date
-        season, _ = ClimaAnalysis.Utils.find_season_and_year(curr_date)
-        push!(seasons, season)
-        curr_date += Dates.Month(3) # Season change every three months
+function find_season_and_year(date::Dates.DateTime)
```
nefrathenrici: Can we use the ClimaAnalysis function instead?

nefrathenrici (approved):
Thanks!

ext/utils.jl
ph-kev: Technically, that function is private since it is not exported. I am fine with using it though since I don't think it will change in the future.

PR #338, opened 2026-07-14
Add asynchronous workers

### LINKED ISSUE #296: Asynchronously request for workers for WorkerBackend
Currently, the `WorkerBackend` request for all the workers at once and wait for all of them to be ready. An alternative to this is to start a `Task` and request individual workers instead of waiting for all of them. This allows work to get started immediately when the workers are available but introduces additional complexity. For example, you can't use `@everywhere` because not all the workers are available.

nefrathenrici:
This PR makes workers individually submitted and self-registering. Each worker is its own scheduler allocation, adds itself to a process-wide pool once it has started and loaded model code, and the scheduler loop dispatches forward-model runs to workers as they become available. Closes #296. 

Changes
- Updated `add_workers` to be asynchronous. This is the new intended way to use the `WorkerBackend`, though `addprocs` still works.
- Add global, self-filling worker pool `GLOBAL_WORKER_POOL`. This pool is guarded by `POOL_LOCK`
- `Distributed.launch` now submits `n` separate single-process allocations instead of one job array. `poll_files_for_worker_startup` now polls all worker signal files, since they spin up separately now.
- Added `@worker_setup` to replace `@everywhere`, which required all workers to be already registered. The new macro records setup expressions on the main process and replays them on every worker as it joins.
- `run_iteration` now starts with a possibly empty pool and waits for workers. Added `EMPTY_POOL_TIMEOUT = 7200` so that jobs don't just hang indefinitely, maybe we can make this configurable.
- Added `cancel_worker_jobs` to cancel all workers. This now runs in an `atexit` hook.

### DIFF
--- a/docs/src/backends.md
+++ b/docs/src/backends.md
@@ -30,10 +30,11 @@ computing environments.
     - [`GCPBackend`](@ref): CliMA's Google cloud platform.
 
 !!! note "What are the differences between the `WorkerBackend` and the HPC cluster backends?"
-    Both distribute ensemble members across a cluster, but in different ways. The
-    `WorkerBackend` keeps a pool of long-lived Julia workers and dispatches
-    forward-model runs to them, so each worker precompiles once and reuses that
-    code across all iterations. The HPC cluster backends instead submit a
+    The main difference between the two backends is how the work of the ensemble
+    members is distributed. The `WorkerBackend` keeps a pool of long-lived Julia
+    workers and dispatches forward-model runs to them, so each worker precompiles
+    once and reuses that code across all iterations. The HPC cluster backends
+    instead submit a
     separate scheduler job for every ensemble member on every iteration; each job
     starts a fresh Julia process and precompiles again, but the jobs are
     independent, so an iteration makes progress as soon as any member is
--- a/docs/src/backends.md
+++ b/docs/src/backends.md
@@ -19,6 +19,9 @@ computing environments.
    or by initializing julia with the `-p` option: `julia -p 2`. Available
    workers can be accessed using
    [`Distributed.workers()`](https://docs.julialang.org/en/v1/stdlib/Distributed/#Distributed.workers).
+   On a cluster, [`add_workers`](@ref) submits each worker as an individual
+   scheduler allocation and adds it to a pool as it connects, so a calibration
+   can start before every worker is up and picks up workers as they join.
 
 3. HPC Cluster Backends: These backends schedule forward model runs on HPC clusters using Slurm or PBS.
     - [`CaltechHPCBackend`](@ref): Caltech's Resnick HPC cluster,
@@ -26,19 +29,25 @@ computing environments.
     - [`DerechoBackend`](@ref): NSF NCAR Derecho supercomputing system,
     - [`GCPBackend`](@ref): CliMA's Google cloud platform.
 
-!!! note "What are the differences between the `WorkerBackend` and the `HPCBackend`?"
-    The `WorkerBackend` assigns work to a pool of long-lived Julia workers, while
-    the `HPCBackend` submits a job for each ensemble member for a single
-    iteration. [`add_workers`](@ref) submits each worker as an individual
-    scheduler allocation, so on a busy cluster the scheduler can start them
-    independently as resources free up rather than needing to place every worker
-    at once. Each worker adds itself to a global pool once it has started and
-    loaded its code, and forward model runs are dispatched to workers as they
-    become available. Use [`@worker_setup`](@ref) instead of `@everywhere` to load
-    model code so that workers joining later are initialized correctly. Because
-    each `HPCBackend` job starts a fresh Julia process, precompilation occurs on
-    every job submission. Workers in the `WorkerBackend` are long-lived, so
-    precompiled code is reused across all iterations.
+!!! note "What are the differences between the `WorkerBackend` and the HPC cluster backends?"
+    Both distribute ensemble members across a cluster, but in different ways. The
+    `WorkerBackend` keeps a pool of long-lived Julia workers and dispatches
+    forward-model runs to them, so each worker precompiles once and reuses that
+    code across all iterations. The HPC cluster backends instead submit a
+    separate scheduler job for every ensemble member on every iteration; each job
+    starts a fresh Julia process and precompiles again, but the jobs are
+    independent, so an iteration makes progress as soon as any member is
+    scheduled. Prefer the `WorkerBackend` when precompilation dominates runtime
+    and the workers fit in your allocation; prefer an HPC cluster backend when
+    forward models are long-running, need internal parallelism, or the cluster
+    cannot hold all workers at once.
+
+!!! note "Loading code on workers"
+    For a `WorkerBackend`, load your model code with [`@worker_setup`](@ref)
+    rather than `Distributed.@everywhere`. Because workers can join
+    asynchronously, `@everywhere` would miss any worker that connects after it
+    runs; `@worker_setup` records the setup and replays it on each worker as it
+    joins, so late-joining workers are initialized correctly.
 
 ## Choosing the right backend for calibration
 
--- a/docs/src/backends.md
+++ b/docs/src/backends.md
@@ -27,16 +27,18 @@ computing environments.
     - [`GCPBackend`](@ref): CliMA's Google cloud platform.
 
 !!! note "What are the differences between the `WorkerBackend` and the `HPCBackend`?"
-    The `WorkerBackend` requests for all the resources up front to do all the
-    work (typically executing a forward model) for a single iteration while the
-    `HPCBackend` submits a job for each ensemble member for a single iteration.
-    On a busy cluster, the `WorkerBackend` can stall an entire iteration waiting
-    for enough free resources to launch all workers at once. The `HPCBackend`
-    avoids this by submitting each ensemble member as an independent job, so
-    progress for a calibration is made as soon as any ensemble member completes.
-    Because each `HPCBackend` job starts a fresh Julia process, precompilation
-    occurs on every job submission. Workers in the `WorkerBackend` are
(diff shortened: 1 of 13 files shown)


docs/src/backends.md
```diff
@@ -27,16 +27,18 @@ computing environments.
...
-    long-lived, so precompiled code is reused across all iterations.
+    The `WorkerBackend` assigns work to a pool of long-lived Julia workers, while
+    the `HPCBackend` submits a job for each ensemble member for a single
+    iteration. [`add_workers`](@ref) submits each worker as an individual
+    scheduler allocation, so on a busy cluster the scheduler can start them
+    independently as resources free up rather than needing to place every worker
+    at once. Each worker adds itself to a global pool once it has started and
+    loaded its code, and forward model runs are dispatched to workers as they
+    become available. Use [`@worker_setup`](@ref) instead of `@everywhere` to load
+    model code so that workers joining later are initialized correctly. Because
```
ph-kev: For this note, I would prefer it talks about the differences between the backends. A lot of content in this note is talking about the HPC Cluster Backends and the relevant functions. Could this be moved into the section for "HPC Cluster Backends"? Maybe it is worth another note about talking about the difference between `@everywhere` and `@worker_setup`.

test/pbs_manager_unit_tests.jl
```diff
@@ -1,9 +1,22 @@
...
 
+# Blocking version of add_workers for test
+function add_workers_and_wait(n; timeout = 1200, kwargs...)
+    ids = fetch(ClimaCalibrate.add_workers(n; kwargs...))
+    pool = ClimaCalibrate.Backend.GLOBAL_WORKER_POOL
+    tstart = time()
+    while !all(id -> id in pool.workers, ids)
+        (time() - tstart) > timeout &&
+            error("Workers did not initialize within $(timeout)s")
+        sleep(0.5)
```
ph-kev, line 12: Is there a better solution than busy waiting? I don't think it need to be addressed in this PR, but there is probably a better solution.

src/backends/workers.jl
```diff
@@ -7,18 +7,280 @@ export add_workers,
...
+    return nothing
+end
+
+"""
+    @worker_setup expr
+
+Like `Distributed.@everywhere`, but the expression is also recorded and replayed
+on any worker that joins later. Use this instead of `@everywhere` to load model
+code for an asynchronous [`WorkerBackend`](@ref) calibration, where workers may
+join after the calibration has started.
```
ph-kev: Can you add more emphasis or use an admonition? I don't think this emphasize the importance of using this macro (e.g. what can happen if you don't?).

src/backends/workers.jl
```diff
@@ -7,18 +7,280 @@ export add_workers,
...
+
+`using`/`import` statements run on the master first (to precompile once), and
+the current source path is propagated so relative `include` works on workers.
+As with `@everywhere`, local variables must be interpolated with `\$`.
```
ph-kev: Maybe I am not too familiar with the teminology, but I don't understand what master is here. Is it the same as the main process or is this specific to the Distributed.jl?

src/backends/workers.jl
```diff
@@ -7,18 +7,280 @@ export add_workers,
...
+# allocations would be orphaned. Every scheduler `launch` registers an `atexit`
+# hook (`cancel_worker_jobs`) that cancels all of this session's jobs by their
+# shared job name.
+# ----------------------------------------------------------------------------
+
+# Ensures the teardown `atexit` hook is registered at most once per session.
+const ATEXIT_HOOK_REGISTERED = Ref(false)
+
+# Run `cmd`, discarding its output.
+_run_quiet(cmd) = run(pipeline(cmd; stdout = devnull, stderr = devnull))
```
ph-kev: When this run, is the output of each of the forward model still saved in the logs?

src/backends/workers.jl
```diff
@@ -7,18 +7,280 @@ export add_workers,
...
+when they start. Used as the default pool for [`WorkerBackend`](@ref).
+"""
+const GLOBAL_WORKER_POOL = WorkerPool()
+
+# Guards mutation of GLOBAL_WORKER_POOL, INITIALIZING, and WORKER_SETUP.
+const POOL_LOCK = ReentrantLock()
+
+# Workers that have connected but are still loading code: present in `workers()`
+# but not yet schedulable. Keeps `default_worker_pool` from pooling them early.
+const INITIALIZING = Set{Int}()
```
ph-kev: This is a nit pick, but I don't like this variable name being a verb or action.

src/backends/workers.jl
```diff
@@ -7,18 +7,280 @@ export add_workers,
...
+
+# Generous bound (seconds) on how long a calibration will wait on an empty
+# worker pool before erroring, so an asynchronous calibration cannot hang
+# forever when no workers ever start.
+const EMPTY_POOL_TIMEOUT = 7200
```
ph-kev: Should this be configurable by the user?

src/backends/workers.jl
```diff
@@ -7,18 +7,280 @@ export add_workers,
...
+
+Cluster workers add themselves via the `:register` hook. Workers added by other
+means (e.g. plain `addprocs`/`LocalManager` or pre-existing workers) are
+reconciled into the pool here. Workers still loading code (in `INITIALIZING`)
+are skipped so they are not scheduled before they are ready.
+"""
+function default_worker_pool()
+    lock(POOL_LOCK) do
+        for id in workers()
+            id == 1 && continue
```
ph-kev: Can you add a comment about the `id == 1` case? I think this is the ID of the main worker, but I am not confident.

src/backends/workers.jl
```diff
@@ -7,18 +7,280 @@ export add_workers,
...
+    end
+    return nothing
+end
+
+"""
+    remove_worker_from_pool(id)
+
+Remove worker `id` from `GLOBAL_WORKER_POOL`. Called when a worker
+deregisters (e.g. walltime expiry or crash). Any stale entry left in the pool's
+channel is filtered out at `take!` time.
```
ph-kev: Can you clarify the last sentence? I am not sure what context this is referring to.

src/calibration.jl
```diff
@@ -403,9 +412,31 @@ function run_iteration(
...
+            if inflight[] > 0 || Backend.n_initializing_workers() > 0
+                t_last_available = time()
+            end
+            t_empty = time() - t_last_available
+            if inflight[] == 0 &&
+               Backend.n_initializing_workers() == 0 &&
+               t_empty > Backend.EMPTY_POOL_TIMEOUT
+                throw(
+                    ErrorException(
+                        "No workers available for $(round(Int, t_empty))s \
```
ph-kev: Is this better than using `error(msg)`?

ph-kev (changes requested):

docs/src/backends.md
nefrathenrici: I changed this to focus on the backend differences

src/backends/workers.jl
nefrathenrici: I added a !!! warning admonition

src/backends/workers.jl
nefrathenrici: yes, i renamed it to the main process

src/backends/workers.jl
nefrathenrici: Yes, it just silences the command (scancel/qdel) output itself.

src/backends/workers.jl
nefrathenrici: Agree, renamed to INITIALIZING_WORKERS

src/backends/workers.jl
nefrathenrici: Yes, added it to the backend

src/backends/workers.jl
nefrathenrici: Yes, it is the ID of the main worker. I added an inline comment

src/backends/workers.jl
nefrathenrici: I expanded it:
```
A `Distributed.WorkerPool` holds its workers both in a `Set` and in an internal
`Channel` that `take!` draws from. This only removes `id` from the `Set`, so a
copy of `id` may still sit in the channel. That copy is harmless because
`WorkerPool`'s `take!` discards any id that is no longer a live process before
returning one.
```

src/calibration.jl
nefrathenrici: No, I changed it to `error`

docs/src/backends.md
```diff
@@ -19,24 +19,35 @@ computing environments.
...
-    `HPCBackend` submits a job for each ensemble member for a single iteration.
-    On a busy cluster, the `WorkerBackend` can stall an entire iteration waiting
-    for enough free resources to launch all workers at once. The `HPCBackend`
-    avoids this by submitting each ensemble member as an independent job, so
-    progress for a calibration is made as soon as any ensemble member completes.
-    Because each `HPCBackend` job starts a fresh Julia process, precompilation
-    occurs on every job submission. Workers in the `WorkerBackend` are
-    long-lived, so precompiled code is reused across all iterations.
+!!! note "What are the differences between the `WorkerBackend` and the HPC cluster backends?"
+    Both distribute ensemble members across a cluster, but in different ways. The
```
ph-kev: For the first sentence, I would write "The main difference between both backends is how the work of the ensemble members are distributed". This current sentence sounds strange to me.

src/backends/workers.jl
```diff
@@ -7,18 +7,285 @@ export add_workers,
...
+    return nothing
+end
+
+"""
+    @worker_setup expr
+
+Like `Distributed.@everywhere`, but the expression is also recorded and replayed
+on any worker that joins later.
+
+!!! warning
```
ph-kev: Maybe the tip admonition would work better here?

ph-kev (approved):
Before merging this, can we test this with a calibration pipeline (assuming if any are using the `WorkerBackend`)?

PR #341, opened 2026-07-30
Add workers_per_node argument to add_workers

nefrathenrici:
Changes
- `add_workers(n; workers_per_node = k)` now packs k workers into a single scheduler allocation (one worker per GPU) instead of one allocation per worker. This is needed for clusters/queues that charge for exclusive use of nodes. Default is workers_per_node = 1 so behavior is unchanged for existing pipelines. Each worker on the same node is pinned to its own GPU by setting CUDA_VISIBLE_DEVICES per worker (see `pbs_multi_worker_script`)
- Worker .out files now default to an auto-cleaned temp dir instead of piling up in the working directory.

### DIFF
--- a/docs/src/backends.md
+++ b/docs/src/backends.md
@@ -22,7 +22,7 @@ computing environments.
    On a cluster, [`add_workers`](@ref) submits each worker as an individual
    scheduler allocation and adds it to a pool as it connects, so a calibration
    can start before every worker is up and picks up workers as they join. On
-   clusters that charge for whole nodes (Derecho), pass `workers_per_node` to
+   clusters that charge for whole nodes (e.g. Derecho), pass `workers_per_node` to
    [`add_workers`](@ref) to put several workers, one per GPU, in each
    allocation.
 
--- a/docs/src/backends.md
+++ b/docs/src/backends.md
@@ -21,7 +21,10 @@ computing environments.
    [`Distributed.workers()`](https://docs.julialang.org/en/v1/stdlib/Distributed/#Distributed.workers).
    On a cluster, [`add_workers`](@ref) submits each worker as an individual
    scheduler allocation and adds it to a pool as it connects, so a calibration
-   can start before every worker is up and picks up workers as they join.
+   can start before every worker is up and picks up workers as they join. On
+   clusters that charge for whole nodes (Derecho), pass `workers_per_node` to
+   [`add_workers`](@ref) to put several workers, one per GPU, in each
+   allocation.
 
 3. HPC Cluster Backends: These backends schedule forward model runs on HPC clusters using Slurm or PBS.
     - [`CaltechHPCBackend`](@ref): Caltech's Resnick HPC cluster,
--- a/src/backends/workers.jl
+++ b/src/backends/workers.jl
@@ -332,12 +332,82 @@ function Distributed.manage(
     return nothing
 end
 
-# Main SlurmManager function, adapted from the unmaintained ClusterManagers.jl
+# Where a worker's startup output goes, default to a temp dir
+function default_worker_output_base(params, exehome, jobname)
+    haskey(params, :o) && return params[:o]
+    haskey(params, :output) && return params[:output]
+    return joinpath(mktempdir(exehome; prefix = ".julia_worker_"), jobname)
+end
+
+"""
+    submit_workers!(instances_arr, launch_condition; single_cmd, multi_worker_cmd, kwargs...)
+
+Submit `ntasks` workers and fill `instances_arr` as they connect. Shared by the
+Slurm and PBS `launch` methods.
+
+When `workers_per_node == 1`, each worker is submitted on its own with
+`single_cmd(output_file)`. Otherwise the workers are grouped into allocations of
+`workers_per_node` and each allocation is submitted with
+`multi_worker_cmd(log_file, script_file)`, where the script runs its workers as
+background processes (see `multi_worker_script`). Worker output files are named
+from `output_base`.
+"""
+function submit_workers!(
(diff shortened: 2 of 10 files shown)


docs/src/backends.md
```diff
@@ -21,7 +21,10 @@ computing environments.
    [`Distributed.workers()`](https://docs.julialang.org/en/v1/stdlib/Distributed/#Distributed.workers).
    On a cluster, [`add_workers`](@ref) submits each worker as an individual
    scheduler allocation and adds it to a pool as it connects, so a calibration
-   can start before every worker is up and picks up workers as they join.
+   can start before every worker is up and picks up workers as they join. On
+   clusters that charge for whole nodes (Derecho), pass `workers_per_node` to
```
ph-kev:
```suggestion
   clusters that charge for whole nodes (e.g. Derecho), pass `workers_per_node` to
```

test/workers_per_node.jl
```diff
@@ -0,0 +1,83 @@
...
+    parse_slurm_worker_params,
+    worker_cookie,
+    PBSManager,
+    SlurmManager
+
+@testset "pbs_multi_worker_script" begin
+    outputs = ["/tmp/w-1.out", "/tmp/w-2.out", "/tmp/w-3.out"]
+    script = pbs_multi_worker_script(
+        "julia",
+        Cmd(["--project=/my proj", "--threads=2"]),
```
ph-kev: Does it matter that it is specifically the `/my` project? I think it would be better to use `@temp` instead (I am also not sure if this command is ran or not).

```suggestion
        Cmd(["--project=@temp proj", "--threads=2"]),
```

src/backends/workers.jl
```diff
@@ -332,12 +332,82 @@ function Distributed.manage(
...
+            push!(pids, open(addenv(cmd, env)))
+            push!(output_files, output_path)
+        end
+    else
+        # Pack several workers into one allocation, one per GPU.
+        counts = workers_per_allocation(ntasks, workers_per_node)
+        njobs = length(counts)
+        for (j, nworkers) in enumerate(counts)
+            worker_outputs =
+                [abspath("$output_base-$j-$g.out") for g in 1:nworkers]
```
ph-kev: I think the if and else conditions should be written similarly. For example, here we use `worker_outputs` but the if condition uses `output_path`. I am also confused why we need to handle the `N = 1` and `N > 1` cases separately?

test/runtests.jl
```diff
@@ -9,6 +9,7 @@ using SafeTestsets
 @safetestset "Julia backend" begin include("julia_backend.jl") end
 @safetestset "HPC config" begin include("backend_config.jl") end
 @safetestset "HPC job scripts" begin include("hpc_job_scripts.jl") end
+@safetestset "Workers per node" begin include("workers_per_node.jl") end
```
ph-kev, line 12: Is this important enough to warrant its own test file?

src/backends/workers.jl
```diff
@@ -810,6 +916,8 @@ end
 function _add_workers(nworkers; device, cluster, time, kwargs...)
     if cluster == :local || (cluster == :auto && !is_cluster_environment())
         @info "Using local processing mode, adding $nworkers worker$(nworkers == 1 ? "" : "s")"
+        get(kwargs, :workers_per_node, 1) > 1 &&
+            throw(ArgumentError("workers_per_node needs a cluster"))
```
ph-kev: I don't think this error message is clear enough. It would be helpful to state what workers_per_node is and why the combination of `workers_per_node > 1` and not having a cluster doesn't make sense.

src/backends/workers.jl
```diff
@@ -740,6 +835,16 @@ default_gpu_kwargs(::PBSManager) = (;
...
 )
 
+# Resources for one allocation of `n` workers. The workers run as `n` background
+# processes in a single task, so that task needs all `n` workers' resources.
+allocation_resource_kwargs(::PBSManager, device, n) = Dict{Symbol, Any}(
+    :l_select => device == :gpu ? "ngpus=$n:ncpus=$(4n)" : "ncpus=$n",
+)
+allocation_resource_kwargs(::SlurmManager, device, n) =
+    device == :gpu ?
+    Dict{Symbol, Any}(:gpus_per_task => n, :cpus_per_task => 4n) :
```
ph-kev, line 844: Why is it `4n` for the `cpus_per_task`? Could you include a comment for this?

ph-kev (changes requested):

src/backends/workers.jl
nefrathenrici: I added a comment: 4 CPUs per GPU, matching the single-worker default ngpus=1:ncpus=4, so n GPU workers need 4n CPUs

src/backends/workers.jl
nefrathenrici: I expanded this error message

test/runtests.jl
nefrathenrici: I didn't see a good file to put these tests into, and I wanted to keep them in runtests since they are cluster-independent. They could live in `hpc_job_scripts.jl` but that is HPCBackend specific, opposite to the WorkerBackend functionality these tests exercise. `slurm/pbs_manager_unit_tests.jl` only run in buildkite, so I can't include them there.

src/backends/workers.jl
nefrathenrici: I renamed output_path/worker_outputs to output/outputs so both branches read the same. I also added a comment explaining the two modes, they need to be separate because multiple workers per allocation require different scheduler commands and default output routing.

ph-kev (approved):

PR #351, opened 2026-08-26
Add utilities for residual

### LINKED ISSUE #350: Add a residual plot
For the AMIP calibration pipeline, we have a plot that looks like

Image

It would be good to implement the bottom row of the plots.

ph-kev:
closes https://github.com/CliMA/ClimaCalibrate.jl/issues/350 - This PR adds a function for computing the residual for any iteration, reconstruct the residual as a `OutputVar` if you are using the ClimaAnalysis extension, and plotting the residual.

This function also fix a bug with the units of reconstructing the diagonal of a covariance matrix.

TODO

- [ ] Check code and docs
- [ ] Add example of how to interpret a plot of the residuals


ext/ClimaCalibrateMakieExt.jl
```diff
@@ -123,6 +130,75 @@ function Makie.plot!(obs_plot::Plot_Obs)
...
+            residual_plot,
+            [-2.0, 2.0],
+            color = :red,
+            linestyle = :dot,
+        )
+    end
+    Makie.scatter!(
+        residual_plot,
+        residual_plot.attributes,
+        1:length(residual),
```
odunbar: Is it useful to have the "x" axis here? Is it really meaningful? Or should we just stack these in a column

src/ekp_utils.jl
```diff
@@ -113,4 +116,57 @@ function get_observations_for_nth_iteration(
...
+    ignore_nan = true,
+)
+    obs_series = EKP.get_observation_series(ekp)
+    obs = get_observations_for_nth_iteration(obs_series, N)
+    stacked_obs = mapreduce(EKP.get_obs, vcat, obs)
+    stacked_noise_var = mapreduce(vcat, obs) do o
+        covs = EKP.get_obs_noise_cov(o, build = false)
+        mapreduce(_diag, vcat, covs)
+    end
+    g_mean = if ignore_nan
```
odunbar: How difficult would it be to have one panel per stacked OutputVar (e.g. per variable?) Is this accessible here?

I wonder if folks might often want to separate this into different variables or different months etc. but as usual this is wrapped up in the metadata information

ext/ClimaCalibrateMakieExt.jl
ph-kev: I am confused by what is being recommended here? Do you want this to be a plot of distribution of values in the residual instead?

src/ekp_utils.jl
ph-kev: If this PR is merged, the user would need to use `ObservationRecipe.reconstruct_residual` to get it into a `OutputVar` and can just plot it using the plotting functions in ClimaAnalysis. I don't think it is worth it to add this kind of customization in ClimaCalibrate.

ext/ClimaCalibrateMakieExt.jl
```diff
@@ -156,6 +232,24 @@ end
...
         return (; title, ylabel, xlabel)
     end
+
+    function Makie.preferred_axis_attributes(
+        ::Type{Makie.Axis},
+        residual_plot::Plot_Residual,
+    )
+        ekp = residual_plot.ekp[]
+        iter = _iter_or_total(ekp, residual_plot.iter[])
+        residual = ClimaCalibrate.residual(
```
nefrathenrici: This is not a blocker, but the residual will be computed twice when calling `plot_residual`, is there a way to pass it to this function?

src/svd_analysis.jl
```diff
@@ -182,6 +182,10 @@ Returns a named tuple with:
   `structured_energy_by_variable` and `residual_norm_by_variable`
 
 Requires ClimaAnalysis to be loaded.
+
+!!! note "Sign convention"
```
nefrathenrici: We should make an issue for this and fix it in a future PR.

nefrathenrici (approved):
Thanks!

ext/ClimaCalibrateMakieExt.jl
odunbar: Given your response below that we cannot do a plot per output var - this comment is redundant. 

I think my only point was that the plot of "y against x" is not meaningful as x is some flattened and stacked object. Instead, yes, one could plot the entire distribution of values for `output_var_1` at `x=1`, the distribution of y for `output_var_2` could be at `x=2`.

Maybe something to develop in future for interpretation.

odunbar (approved):
LGTM

In future, we can think about having a more interpretable "x" axis - but this is already a sufficient improvement for now

src/ekp_utils.jl
```diff
@@ -180,4 +183,57 @@ function get_observations_for_nth_iteration(
...
+    ekp::EKP.EnsembleKalmanProcess;
+    N = EKP.get_N_iterations(ekp),
+    ignore_nan = true,
+)
+    obs_series = EKP.get_observation_series(ekp)
+    obs = get_observations_for_nth_iteration(obs_series, N)
+    stacked_obs = mapreduce(EKP.get_obs, vcat, obs)
+    stacked_noise_var = mapreduce(vcat, obs) do o
+        covs = EKP.get_obs_noise_cov(o, build = false)
+        mapreduce(_diag, vcat, covs)
```
ph-kev: I need to check if this work with an `UniformScaling` matrix.

ph-kev: It looks like this is not a problem.

nefrathenrici (approved):

